PR #1, opened 2014-06-30
[PORT] prepare port to 8.0 by disabling all module

yvaucher:


pedrobaeza:
Don't we agree to put all modules in a folder called **unported**?

yvaucher:
I followed instruction on wiki:

```
It's recommended to rename or move the the modules to more easily identify the unported ones.
This script renames the modules with a _unported suffix so they won't be tested by Travis:

find . -mindepth 1  -maxdepth 1 -type d -not -path '*/\.*'  -execdir git mv {} {}_unported \;
```

pedrobaeza:
Hi, Yannick, due to the problems exposed on the mailing list about this approach, I recommend you to change to the other option. We will change documentation soon to reflect the new approach.

PR #6, opened 2014-07-07
[TEST] DON'T MERGE 7.0 travis coverage encaps test

yvaucher:


.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+ - travis_install_nightly 7.0
+ - pip install coveralls flake8
+ - pip install PyPDF2==1.18
+ - git clone https://github.com/OCA/reporting-engine -b 7.0 $HOME/reporting-engine
+ - git clone https://github.com/OCA/stock-logistics-workflow -b 7.0 $HOME/stock-logistics-workflow
+ - git clone https://github.com/OCA/webkit-tools -b 7.0 $HOME/webkit-tools
+
+script:
+    - travis_run_flake8
+    - travis_run_tests 7.0 openerp_test $HOME/reporting-engine $HOME/stock-logistics-workflow $HOME/webkit-tools
```
eLBati: I think you should use the travis_run_tests_encapsulated script, if you want to test one module at a time

bwrsandman:
Shouldn't it be a red flag that two modules are incompatible?

yvaucher:
@bwrsandman incompatible tests don't mean incompatible modules.

Just imagine a module adding a required field and an other module without dependency not knowing about this field but using same model. Second module tests will fail when trying to create this kind of object.

yvaucher:
@eLBati I forgot to RTFM :sweat_smile:

guewen:
@bwrsandman I don't know what happens here (didn't look), but 2 modules may have incompatible **tests** whenever the modules are compatible. See my example: https://lists.launchpad.net/openerp-community/msg06379.html

eLBati:
@bwrsandman also see http://openerp-community.2306076.n4.nabble.com/Openerp-community-OCA-transition-from-Launchpad-to-Github-tp4645893p4646141.html

bwrsandman:
Makes sense, but let's be vigilant for incompatible modules.

Adding a required field **can** cause automated behaviour, as well as tests to fail. Testing independently won't catch this.

A way around this is to make "glue" modules. Make it depend on both modules and and have `auto_install = True`, then fix the incompatibility in both the tests and the behaviour.

eLBati:
@bwrsandman I agree: incompatible modules should be tracked.

But module B can't predict every possible use case generated by installing other modules. Module B could even ignore their existence. And it could happen that they will never be installed together in real use cases.
This is why I said that for every module, the significant test is that which tests
the single module.

eLBati:
@yvaucher we got a false positive
https://travis-ci.org/OCA/account-invoice-reporting/builds/29325041#L491
:grimacing:

bwrsandman:
@nbessi ping

yvaucher:
@eLBati same player play again. I used the cheat code @nbessi this time

eLBati:
:worried: 
do you have any suggestion about how to reproduce the error locally?

Meanwhile, let's try with a more verbose test:
https://github.com/yvaucher/account-invoice-reporting/pull/1/files

eLBati:
The test fails because it expects to get
`<ul> <li>S/N Lot0 for Ice cream</li> </ul>`
but it gets
`<ul><li>S/N Lot0 for Ice cream</li> </ul>`

The code generating that note is

```
    def load_line_lots(self, cr, uid, ids, context=None):
        for line in self.browse(cr, uid, ids, context):
            if line.prod_lot_ids:
                note = u'<ul> '
                note += u' '.join([
                    u'<li>S/N {0}</li>'.format(lot.name)
                    for lot in line.prod_lot_ids
                ])
                note += u' </ul>'
                line.write({'formatted_note': note}, context=context)
        return True
```

so the first result seems correct, and it's what I get locally.
I don't know why on travis the space after `<ul>` is removed.
Any clue?

yvaucher:
Great it fixed the issue only PEP8 spam now.

yvaucher:
I close this one. encapsulated test seems unnecessary

eLBati:
@yvaucher Ok.
In general, for the reasons described here https://github.com/OCA/account-invoice-reporting/pull/6#issuecomment-48189581 , I think it's always better to test modules individually, instead of all together.

yvaucher:
@eLBati I agree on this but as tests pass I think we can wait on nbessi code to be merged in OCA maintainer-quality-tools

bwrsandman:
@eLBati how about both? :)
Just add another line in `script`

eLBati:
@yvaucher :ok_hand:
@bwrsandman it can certainly be useful to see both results

PR #13, opened 2014-10-09
[IMP] [7.0] PEP8

eLBati:


pedrobaeza:
There's a problem with missing security rules, but PEP8 is OK, so :+1:

bwrsandman:
I suggest fixing the security rule, the tests will be green after that.

tafaRU:
PEP8 is OK, so one more :+1:

tafaRU:
The security rule needs to be fixed in https://github.com/OCA/reporting-engine/tree/7.0/base_report_assembler So I think this PR can me merged as is.

yvaucher:
:+1:

PR #15, opened 2014-10-21
Extracted base_condition_template from sale_condition_template

guewen:
Extracted from https://github.com/OCA/sale-workflow/pull/50

### DIFF
--- a/base_condition_template/condition.py
+++ b/base_condition_template/condition.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+#
+#
+#    Author: Nicolas Bessi
+#    Copyright 2013-2014 Camptocamp SA
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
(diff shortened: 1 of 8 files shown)


yvaucher:
perfect :+1: 

Thanks

base_condition_template/condition.py
```diff
@@ -0,0 +1,42 @@
...
+class BaseConditionTemplate(models.Model):
+    _name = "base.condition.template"
+    _description = "Base condition template text"
+
+    name = fields.Char('Condition summary', required=True)
+    position = fields.Selection([('before_lines', 'Before lines'),
+                                 ('after_lines', 'After lines')],
+                                'Position',
+                                required=True,
+                                help="Position on document")
```
pedrobaeza, line 34: For required fields, you should provide a default value.

yvaucher, line 34: @pedrobaeza What if the field must be defined ?

Takes account_invoice partner_id is not defined by default.

pedrobaeza, line 34: You're right, but in this case, I think we can default to one of them.

guewen, line 34: Changed

pedrobaeza:
Thanks :+1:

PR #25, opened 2015-05-05
[ADD] Add new module invoice_picking_address_template

hurrinico:


invoice_report_picking_address/README.rst
```diff
@@ -0,0 +1,11 @@
+Invoice Report Picking Address
+==============================
+
+Adds delivery address on account invoice template
```
eLBati: I would replace ```template``` with ```report```

eLBati:
I think module https://github.com/OCA/account-invoice-reporting/tree/7.0/account_invoice_delivery_address should be ported to 8.0 instead of creating this one

eLBati:
@hurrinico what do you think?

eLBati:
I noticed the old module had to change RML report and used a different approach about `partner_shipping_id` field.
I am :+1: with this module, but I would name it `account_invoice_picking_address` and add the dependency on sale_stock, adding the ability to retrieve the address from SO

tafaRU:
@hurrinico, the git author of you commit seems unknown, could you please check it?

eLBati:
Closing this in favour of https://github.com/OCA/account-invoice-reporting/pull/40

PR #14, opened 2014-10-21
Add invoice condition template

guewen:
Display the condition templates on the invoice report.

Needs:
- [x] https://github.com/OCA/account-invoice-reporting/pull/15

PR #26, opened 2015-06-09
[MIG] invoice_webkit and account_draft_invoice_print

charbeljc:


pedrobaeza:
Please rename invoice_webkit to account_invoice_webkit.

yvaucher:
@pedrobaeza we already accepted https://github.com/OCA/account-invoice-reporting/pull/14

I don't think account_invoice_\* has more value than invoice_*

pedrobaeza:
My reasoning is to always put area/application as first word. If we couldn't do it in the other module, we can still do it in this one.

eLBati:
Tried on
http://3112963-26-4786b0.runbot1.odoo-community.org (**_all** DB)
open SAJ/2015/004 and clicked on 'print'.

Get

> Bad Report Reference
> This report is not loaded into the database: account.report_invoice

feketemihai:
@pedrobaeza I think this one should be closed as well since a lot of time without activity.

PR #24, opened 2015-03-16
[7.0] Partner Aging

bwrsandman:
Port of partner_aging from lp:~rcarnes/openerp-shared/7.0

PR #27, opened 2015-07-06
Account payment cc

bwrsandman:
Split of #24


eLBati:
It seems this PR has wrong target

bwrsandman:
Where should it go?

pedrobaeza:
In this case, it seems a mix of account-payment and HR stuff. I would close it and let the original authors to work on it if interested.

bwrsandman:
@max3903 what do you think?

max3903:
:+1: to move it to account_payment

PR #28, opened 2015-07-06
Account deposit

bwrsandman:
Split of #24


account_deposit/__openerp__.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+
+{
+    "name": 'Customer Deposits',
+    "version": '1.7',
+    "author": 'Ursa & OpenERP SA',
```
gurneyalex, line 6: please add OCA

gurneyalex:
not sure this is the correct repository. has it got anything to do with invoice reporting?

PR #30, opened 2015-07-06
L10n us administrative

bwrsandman:
Split of #24


max3903:
Should be moveed to https://github.com/OCA/l10n-usa

PR #38, opened 2015-08-13
8.0 account draft invoice print

damdam-s:


pedrobaeza:
I don't understand the need of this module. You can already print draft invoices. What is the use case?

damdam-s:
@pedrobaeza this module add the button print when invoice state is draft.
If this module is not installed, you don't have the button in the form view.

I agree that printing is allowed with dropdown menu "Print" but having a button is more usable for some users.

gurneyalex:
@damdam-s does not work on draft supplier invoices.

damdam-s:
@gurneyalex no buttons in supplier view on standard

pedrobaeza:
I will let then clear the purpose on the README file (the same as you have explained me). I still don't see much utility for this module (it's only 1 click more), but if others want to have it...

gurneyalex:
@damdam-s you're right. Sorry for the noise

@pedrobaeza my users generally find very unfriendly the 2 dropdown lists for reports and 'more actions' in the form view, and I tend to agree that on a daily basis it really sucks. Regarding this specific button, lots of customers are still "paper based" and proofreading an invoice formatted in the way you're used to, without having to check information in various tabs is something people who manually input invoices do every day. Even if the invoice is not printed only read on screen in a PDF viewer, you get mistakes undetectable in the form view which jumps to your eyes (wrong partner, wrong taxes, missing payment terms...). 

And one click avoided 10 times / day is something which gets you small cakes and a smile from the customers Well... actually no, you don't get this, unless you let them work with the extra click for some time first :smile_cat:

pedrobaeza:
OK then, but please explain the purpose and extract the README according OCA's README template.

damdam-s:
done

account_draft_invoice_print/README.rst
```diff
@@ -0,0 +1,42 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Account Draft Invoice Print
+===========================
+
+This module add the button print when invoice's state is draft.
```
pedrobaeza: s/add/adds

pedrobaeza: s/invoice/customer invoice

damdam-s:
update

LeartS:
I'm also unsure whether the change deserves a module, but this repo has only 2 modules at the moment, so I think it's ok.
LGTM :+1:

yvaucher:
:+1:

account_draft_invoice_print/__openerp__.py
```diff
@@ -22,12 +22,11 @@
 {'name': 'Account Draft Invoice Print',
  'version': '1.0.0',
```
yvaucher: Please use full version number

https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

damdam-s:
done

account_draft_invoice_print/README.rst
```diff
@@ -0,0 +1,42 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Account Draft Invoice Print
```
angelmoya, line 6: Set upper line and agpl target from as in new template
 https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

rest is ok for me :+1:

damdam-s, line 6: @amoya-dx : done. waiting for green ...

damdam-s:
updated

angelmoya:
:+1:

PR #50, opened 2017-01-11
[ADD] account_reporting_weight

pedrobaeza:
Weights in the invoices analysis view
=====================================

This module adds the measure "Weight" in the invoices analysis view. This is
caught from 2 possible sources:

* If the UoM of the product is one of the category "Weight", the value is taken from the ordered quantity.
* If the UoM of the product is another, then the weight is taken from the weight field of the product multiply by the ordered quantity.

Configuration
=============

You need to be at least "Accountant" on "Accounting & Finance" role for seeing the report.

Usage
=====

* Go to *Invoicing > Reporting > Businness Intelligence > Invoices*.
* Add the "Weight" measure from your "Measures" dropdown in your analysis.

Known issues / Roadmap
======================

* The weight quantity is expressed in the unit of measure of the product, so if you have several weight UoMs across your products, the global sum won't make sense.

### DIFF
--- a/account_reporting_weight/README.rst
+++ b/account_reporting_weight/README.rst
@@ -1,0 +1,61 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================================
+Weights in the invoices analysis view
+=====================================
+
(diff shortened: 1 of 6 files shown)


pedrobaeza:
cc @Tecnativa

rafaelbn:
@Tecnativa

luismontalba (approved):
Functionality tested on runbot

pedrobaeza:
OK, this one is good to merge.

Issue #52, opened 2017-04-14
partner_daytopay


max3903:
At @ursais, we have a module to compute some financial statistics for a partner:
* Average number of days to get customer invoices paid (in the current year)
* Average number of days to get customer invoices paid (lifetime)
* Average number of days to pay vendor bills (in the current year)
* Average number of days to pay vendor bills (lifetime)

I think this repository would be the best fit to host it, unless the feature already exists in a module I haven't found.

Any opinion? Interest?

Thanks.

JordiBForgeFlow:
I find the idea interesting. I tend to think that https://github.com/OCA/account-financial-tools would be better. This is not really a report but a KPI.

The link is broken, though.

max3903:
@jbeficent Sorry, private repo... I removed the link.

PR #53, opened 2017-04-20
[ADD] partner_daytopay

max3903:
This module displays statistics related to the receivables and payables behavior of a partner on the Accounting tab of the partner form view.

# Usage

* Go to the Accounting tab of a partner
* There is 4 new fields. These fields represent the average days to pay for this year (YTD) and since the first invoice (lifetime).

#52


partner_daytopay/README.rst
```diff
@@ -0,0 +1,75 @@
...
+   :alt: License: AGPL-3
+
+===================
+Partner Days to Pay
+===================
+
+This module displays statistics related to the receivables and payables behavior of a partner on the Accounting tab of the partner form view.
+
+Installation
+============
```
lasley: Can remove sections without specific instructions

partner_daytopay/models/res_partner.py
```diff
@@ -0,0 +1,158 @@
...
+            invoice_ids = partner.env['account.invoice'].search(invoice_domain)
+            for invoice in invoice_ids:
+                total_number_of_invoices += 1
+                date_due = invoice.date_invoice
+                days_for_latest_payment = 0
+
+                for payment in invoice.payment_ids:
+                    if payment.state == 'posted':
+                        days_for_this_payment = \
+                            (datetime.strptime(payment.payment_date,
```
lasley: `fields.Datetime.from_string` instead of datetime

partner_daytopay/models/res_partner.py
```diff
@@ -0,0 +1,158 @@
...
+                                days_for_latest_payment = days_for_this_payment
+
+                    days_to_pay_invoice = days_for_latest_payment
+                    total_days_to_pay = total_days_to_pay + days_to_pay_invoice
+                    average_days_to_pay = total_days_to_pay / \
+                        total_number_of_invoices
+
+            partner.d2r_ytd = average_days_to_pay
+
+    d2p_life = fields.Float(compute=_compute_dtp_life,
```
lasley: Move field declarations to top

partner_daytopay/models/res_partner.py
```diff
@@ -0,0 +1,158 @@
...
+
+        for partner in self:
+            average_days_to_pay = 0
+            total_days_to_pay = 0
+            total_number_of_invoices = 0
+            invoice_domain = [('partner_id', '=', partner.id),
+                              ('state', '=', 'paid'),
+                              ('type', '=', 'out_invoice')]
+            invoice_ids = partner.env['account.invoice'].search(invoice_domain)
+            for invoice in invoice_ids:
```
lasley: These computation methods are duplicating a lot of logic and should be unified or abstracted

partner_daytopay/models/res_partner.py
```diff
@@ -0,0 +1,158 @@
...
+                    days_to_pay_invoice = days_for_latest_payment
+                    total_days_to_pay = total_days_to_pay + days_to_pay_invoice
+                    average_days_to_pay = total_days_to_pay / \
+                        total_number_of_invoices
+
+            partner.d2r_ytd = average_days_to_pay
+
+    d2p_life = fields.Float(compute=_compute_dtp_life,
+                            string='AVG Days to Pay (lifetime)')
+    d2p_ytd = fields.Float(compute=_compute_dtp_ytd,
```
lasley: Use strings for compute method

lasley (changes requested):
Thanks @max3903 - comments inline

partner_daytopay/models/res_partner.py
max3903: @bodedra Can you help me on this? Thanks!

partner_daytopay/views/res_partner.xml
```diff
@@ -0,0 +1,31 @@
...
+                    <field name="d2p_life"/>
+                </field>
+
+                <field name="property_supplier_payment_term_id" position="after">
+                    <field name="d2r_ytd"/>
+                    <field name="d2r_life"/>
+                </field>
+
+                <field name="credit" position="replace">
+                    <field name="credit" modifiers="{'readonly': true}"/>
```
jcdrubay: Why do you need this? This is a function field without inverse right and therefore already readonly, right? Maybe we could have a comment to explain why in the code?

partner_daytopay/models/res_partner.py
```diff
@@ -0,0 +1,83 @@
...
+            ('type', '=', type)
+        ])
+
+    @api.multi
+    def _get_avg_pay_days(self, invoice_ids):
+        average_days_to_pay = 0
+        total_days_to_pay = 0
+        total_number_of_invoices = 0
+        for invoice in invoice_ids:
+            if fields.Date.from_string(invoice.date_invoice).year ==\
```
jcdrubay: With this if, it will always be YTD and never lifetime.

partner_daytopay/models/res_partner.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @api.multi
+    def _compute_dtp_life(self):
+
+        for partner in self:
+            invoice_ids = self._get_invoice_ids(partner.id, 'out_invoice')
+            partner.d2p_life = self._get_avg_pay_days(invoice_ids)
+
+    @api.multi
+    def _compute_dtp_ytd(self):
```
jcdrubay: Exactly the same code as _compute_dtp_life

jcdrubay (changes requested):
I don't understand how this code works for lifetime; it seems to me that it will always only compute ytd indicators.

Maybe basic automatic tests would highlight this with 3 invoices:
 invoice 1 for partner 1 this year
 invoice 2 for partner 2 this year
 invoice 3 for partner 3 last year

jcdrubay:
You could create a unique compute function for all fields `_compute_d2x`

Also, I would refactor so that we only loop once per invoice type. Here is a suggestion...

```
    def _compute_d2x(self):
        for partner in self:
            partner.d2p_ytd, partner.d2p_life = \
                self._compute_d2x_per_invoice_type(partner, 'out_invoice')
            partner.d2r_ytd, partner.d2r_life = \
                self._compute_d2x_per_invoice_type(partner, 'in_invoice')

    def _compute_d2x_per_invoice_type(self):

        this_year = datetime.now().year
    
        total_number_of_invoices_life = 0
        total_number_of_invoices_ytd = 0

        total_days_to_pay_life = 0
        total_days_to_pay_ytd = 0

        for invoice in self._get_invoice_ids(partner.id, invoice_type):

            invoice_year = fields.Date.from_string(invoice.date_invoice).year
 
            date_due = fields.Date.from_string(invoice.date_invoice)

            days_to_pay_invoice = self._get_invoice_payment(
                    invoice.payment_ids, date_due)

            total_number_of_invoices_life += 1
            total_days_to_pay_life += days_to_pay_invoice
 
            if invoice_year == this_year:
                total_number_of_invoices_ytd += 1 
                total_days_to_pay_ytd += days_to_pay_invoice

        d2x_ytd = days_to_pay_ytd / number_of_invoices_ytd
        d2x_life = days_to_pay_life / number_of_invoices_life
        
        return d2x_ytd, d2x_life
```

partner_daytopay/models/res_partner.py
```diff
@@ -0,0 +1,82 @@
...
+            days_to_pay_invoice = self._get_invoice_payment(
+                    invoice.payment_ids, date_due)
+
+            total_number_of_invoices_life += 1
+            total_days_to_pay_life += days_to_pay_invoice
+ 
+            if invoice_year == this_year:
+                total_number_of_invoices_ytd += 1 
+                total_days_to_pay_ytd += days_to_pay_invoice
+                d2x_ytd = total_days_to_pay_ytd / total_number_of_invoices_ytd
```
jcdrubay: This should be outside the for loop because we only need to calculate once.

partner_daytopay/models/res_partner.py
```diff
@@ -0,0 +1,82 @@
...
+            
+            d2x_life = total_days_to_pay_life / total_number_of_invoices_life
```
jcdrubay: This should be outside the for loop because we only need to calculate once.

jcdrubay (changes requested):
Minor optimization suggestion

partner_daytopay/models/res_partner.py
jcdrubay: and we should have a little check to avoid division by 0 after bringing it outside the loop :)

partner_daytopay/models/res_partner.py
jcdrubay: and we should have a little check to avoid division by 0 after bringing it outside the loop :)

jcdrubay (changes requested):

vprime (approved):
Looks good

jcdrubay:
@max3903 I am surprised that you merged without 3 approvals. Also I am not sure if you have seen my last 2 comments. If you did then I guess I was  mistaken. Anyway, I think that it is a nice contribution :)

max3903:
@jcdrubay I am sorry I didn't see your comments. Fixed in #59. Yes I merged without 2 approvals because this PR had been opened for almost a month...

PR #55, opened 2017-05-16
[FIX] typo into view files definition

vrenaville:
Hello,

This module cannot be installed due to typo in definitions of file into __openerp__,py,


yvaucher (approved):

yvaucher:
Fast tracked as obvious error

PR #59, opened 2017-06-26
[IMP] Catch ZeroDivisionError

max3903:


jcdrubay:
Thanks for your additional commit. 
I am sorry but I don't understand why you don't bring the code of this commit outside the for loop. 

BTW I don't think you can get a zeroDivisionError if you keep it inside the for loop.

lasley:
Hey @max3903 - Did you actually run into this error or something? How can the total number of invoices be less than zero?

max3903:
@lasley Not I didn't, but it can be 0.

lasley (approved):

jcdrubay (approved):

elicoidal (approved):

PR #56, opened 2017-05-25
[MIG][10.0] account_invoice_report_by_partner module

cubells:
- Update manifest file
- Update views
- Update tests

cc @Tecnativa

PR #60, opened 2017-08-26
[10.0][MIG] Migrate account_invoice_report_by_partner

robyf70:
Patch to migrate to v10.0


pedrobaeza:
Thanks for the contribution, but there's already another migration on progress on #56. Please review that one instead.

PR #62, opened 2017-09-15
[FIX] Search for partner or its parent to get invoice ids

max3903:
The module needs to include invoices sent to either partner_id or invoice address child of the partner_id.


max3903:
:+1:

feketemihai (approved):

PR #66, opened 2017-11-02
[ADD] base_comment_template: migration scripts

cubells:
Script necessary if we do migration from v7.0 of base_comment_template module

cc @Tecnativa


pedrobaeza:
Check Travis errors (not related to your change but needed).

base_comment_template/migrations/8.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,20 @@
...
+    (
+        'base_condition_template',
+        'base_comment_template'
+    ),
+]
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    if not version:
```
pedrobaeza: This is not needed it you put `@openupgrade.migrate()`

pedrobaeza (approved):
We can proceed with the merge of this specific case.

Issue #63, opened 2017-10-03
Migration to version 11.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

# Modules to migrate

- [ ] account_draft_invoice_print
- [x] account_invoice_line_report - By @cubells - #69 
- [x] account_invoice_production_lot - By @pedrobaeza - #95
- [ ] account_invoice_report_by_partner
- [x] account_invoice_report_due_list - By @cubells - #94 
- [x] account_reporting_weight - By @chienandalu - #90 
- [x] base_comment_template - By @xavierjimenez - #80 
- [ ] html_invoice_product_note
- [x] invoice_comment_template > account_invoice_comment_template - By @xavierjimenez - #81 
- [ ] invoice_print_report_balance_payment
- [ ] invoice_report_assemble
- [ ] invoice_webkit
- [x] partner_time_to_pay by @max3903 in #67

pedrobaeza:
Migrating `account_invoice_production_lot`

emagdalenaC2i:
@pedrobaeza Please add to the list invoice_comment_template that is renamed as account_invoice_comment_template and merged in #81 
add to the list account_invoice_report_due_list that is migrated and merged in #94

PR #67, opened 2017-12-05
[MIG] partner_time_to_pay

max3903:
#63


lasley:
Hmmm I wonder if this should be renamed to `partner_day_to_pay`?

partner_daytopay/README.rst
```diff
@@ -0,0 +1,66 @@
...
+=====
+
+* Go to the Accounting tab of a partner
+* There is 4 new fields. These fields represent the average days to pay for
+  this year (YTD) and since the first invoice (lifetime).
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/94/11.0
+
```
feketemihai: @max3903 Can you add in roadmap, that for payment terms with multiple payment lines, it will get the value from the date_due of the invoice, it will not threat every payment line separately, because you can have delays on first payments, but the last one to be at term, and this will not show the real delays..

partner_daytopay/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Partner Days to Pay',
+    'summary': 'Add receivables and payables statistics to partners',
+    'version': '11.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Open Source Integrators, Odoo Community Association (OCA)',
+    'category': 'Accounting & Finance',
+    'website': 'http://www.opensourceintegrators.com',
```
RoelAdriaans: As per new __manifest__ , can you change the information, and website?
See https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py

RoelAdriaans (changes requested):

pedrobaeza:
Shouldn't this go into account-payment, not here?

I want to add partner holidays management (non payment intervals) to this module, so a more general name like `partner_payment_term` could be great

krijn1981 (approved):
Functionally tested okay

RoelAdriaans:
@pedrobaeza I think a case can be made for both options...
This is used for invoice reporting, and it's payment related.. 

partner_payment_term sound more like an module to extend the partner with a payment term field, that's already there ofcourse.. :)

pedrobaeza:
@RoelAdriaans yeah, you're right. It can fit also for the holidays (non payment intervals).

pedrobaeza:
But the proposed one should be done at least: `partner_day_to_pay`. The `day` part don't convince me yet, but I don't have a better name.

max3903:
@pedrobaeza What about `partner_time_to_pay`?

pedrobaeza:
Yeah, or `partner_date_to_pay`

RoelAdriaans (approved):

PR #69, opened 2018-03-20
[MIG] account_invoice_line_report: Migration to 11.0

cubells:
- Update manifest
- Update views
- Update translation

cc @Tecnativa


account_invoice_line_report/i18n/account_invoice_line_report.pot
```diff
@@ -0,0 +1,53 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove POT file.

account_invoice_line_report/i18n/antic.po
```diff
@@ -0,0 +1,61 @@
+# Translation of Odoo Server.
```
pedrobaeza: Why this language name?

pedrobaeza (changes requested):

feketemihai:
@pedrobaeza Do you use this one, why not using the invoice reporting because it's already based on the invoice lines, just you don't have the filter without price? Is it mandatory in your workflow?

pedrobaeza:
Yes, it's not enough for making some aggregation and it's simpler to have the list view in this case. We have several customers using this directly.

feketemihai (approved):
Ok... Thanks for the feedback.

pedrobaeza (approved):

PR #61, opened 2017-09-08
[MIG] account_invoice_production_lot: Migration to 10.0

cubells:
* Update README
* Update views
* Update translations
* Update demo data
* Update tests

cc @Tecnativa

PR #70, opened 2018-04-04
[MIG] account_invoice_production_lot: Migration to 9.0

mehdi-ghezal:
Update manifest
Update demo data
Update model
Update views
Update tests


pedrobaeza:
Why do you remove demo data? I think it's better to base your migration in the already done 10.0: https://github.com/OCA/account-invoice-reporting/pull/61

mehdi-ghezal:
Did I miss something because it's seems to me that I didn't remove demo data, I just migrate them to XML format.

pedrobaeza:
Sorry, I didn't see the XML file. It's OK then.

PR #71, opened 2018-04-04
[ADD] invoice_comment_template: add template to partner

cubells:
Add missing feature from older version: now you can assign a comment template to partner.

- Update spanish translation

cc @Tecnativa


pedrobaeza:
invoice_comment_template? Why don't we called it `account_invoice_comment_template`?

invoice_comment_template/models/account_invoice.py
```diff
@@ -27,3 +27,13 @@ def _set_note2(self):
         comment = self.comment_template2_id
         if comment:
             self.note2 = comment.get_value(self.partner_id.id)
+
+    @api.onchange('partner_id', 'company_id')
+    def _onchange_partner_id(self):
+        res = super(AccountInvoice, self)._onchange_partner_id()
+        comment_template = self.partner_id.comment_template_id
+        if comment_template.position == 'before_lines':
```
pedrobaeza, line 36: Don't perform anything if not `comment_template`

invoice_comment_template/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
...
+# Copyright 2018 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    comment_template_id = fields.Many2one(
```
pedrobaeza, line 11: Add this field to commercial fields.

pedrobaeza (changes requested):

cubells:
invoice_comment_template is the original module name @pedrobaeza 
Do you think I must change it?

pedrobaeza:
Well, not now. We should do it before, but now it's late. On next version...

invoice_comment_template/models/res_partner.py
cubells: added.

invoice_comment_template/models/account_invoice.py
cubells: It's removed template_id if there is not not comment_template

cubells:
@pedrobaeza can you recheck?

pedrobaeza (approved):

PR #72, opened 2018-04-05
[MIG] account_invoice_report_grouped_by_picking: Migration to 11.0

chienandalu:
Account Invoice Group Picking
=============================

This module allows print invoices with picking info. This module groups
invoice lines and shows information related to sales and pickings on every
group.

cc @Tecnativa


account_invoice_report_grouped_by_picking/views/report_invoice.xml
```diff
@@ -0,0 +1,67 @@
...
+            <t t-set="line_picking_subtotal" t-value="line_picking_qty * (l.price_subtotal / l.quantity)"/>
+            <span t-esc="line_picking_subtotal"
+                  t-options='{"widget": "monetary", "display_currency": "o.currency_id"}'/>
+            <t t-set="subtotal" t-value="(subtotal or 0.0) + line_picking_subtotal"/>
+        </t>
+    </xpath>
+    <xpath expr="//td/span[@t-field='l.price_subtotal']" position="attributes">
+        <attribute name="t-if">not picking</attribute>
+    </xpath>
+    <xpath expr="//tbody[@class='invoice_tbody']" position="inside">
```
pedrobaeza: This block has to be removed:

* The position is very relaxed, being placed when multiple inheritance randomly.
* You shouldn't see latest picking subtotal this way.

account_invoice_report_grouped_by_picking/views/report_invoice.xml
```diff
@@ -0,0 +1,67 @@
...
+                </td>
+            </tr>
+            <t t-set="last_picking" t-value="picking"/>
+        </t>
+    </xpath>
+    <xpath expr="//td/span[@t-field='l.quantity']" position="before">
+        <t t-if="picking">
+            <!-- Compute product qty for that picking and invoice line -->
+            <t t-set="line_picking_qty"
+               t-value="sum(l.mapped('move_line_ids').filtered(lambda x: x.picking_id == picking).mapped('product_uom_qty'))"/>
```
pedrobaeza: Don't use picking qtys, but invoiced qtys.

account_invoice_report_grouped_by_picking/views/report_invoice.xml
```diff
@@ -0,0 +1,67 @@
...
+        <t t-set="subtotal" t-value="0.0"/>
+    </xpath>
+    <xpath expr="//tr[@t-foreach='o.invoice_line_ids']" position="attributes">
+        <attribute name="t-foreach">o.lines_grouped_by_picking()</attribute>
+        <attribute name="t-as">lines_group</attribute>
+    </xpath>
+    <xpath expr="//td/span[@t-field='l.name']/.." position="before">
+        <t t-set="l" t-value="lines_group['line']"/>
+        <t t-set="picking" t-value="lines_group['picking']"/>
+        <t t-if="picking != last_picking">
```
pedrobaeza: This should work for all the pickings including last, so you need to change the logic and put it at the end instead at the beginning.

pedrobaeza (changes requested):

account_invoice_report_grouped_by_picking/views/report_invoice.xml
chienandalu: Doesn't it? All the moves considered are the ones related to the current line and picking.

pedrobaeza: No, you can invoice less quantity than the delivered. I understand this subtotal as the qty invoiced related to that picking, not the delivered one.

chienandalu:
@pedrobaeza Changes done

pedrobaeza (approved):

pedrobaeza (approved):

rafaelbn (approved):
Tested functionally :+1:

pedrobaeza:
Please squash before merging

hieulucky111 (approved):

PR #84, opened 2018-10-09
[10.0][FIX] account_invoice_production_lot: only compute stock.production.lot related to move done

tafaRU:
This patch fixes the following use case scenario.

Steps to reproduce:

* create a new product 'Product Test': stockable and tracking by lots
* update its quantity to 3 with new lot 'Lot1'
* update its quantity to 3 with new lot 'Lot2'
* sale 4 of 'Product Test'
* on the delivery, split the lots and set To Do quantities as 3 for Lot1 and leave 0 for Lot2
* validate the delivery and create a backorder

Current behavior:

Expected behavior:

the invoice report only contains info related to the lot involved in the delivery, which is Lot1.

### DIFF
--- a/account_invoice_production_lot/models/account_invoice.py
+++ b/account_invoice_production_lot/models/account_invoice.py
@@ -36,7 +36,8 @@ class AccountInvoiceLine(models.Model):
             if not line.order_line_ids:
                 return
             line.prod_lot_ids = self.mapped(
-                'order_line_ids.procurement_ids.move_ids.lot_ids')
+                'order_line_ids.procurement_ids.move_ids').filtered(
+                lambda move: move.state == 'done').lot_ids
 
     @api.multi
     def _compute_line_lots(self):
--- a/account_invoice_production_lot/models/account_invoice.py
+++ b/account_invoice_production_lot/models/account_invoice.py
@@ -35,9 +35,11 @@ class AccountInvoiceLine(models.Model):
         for line in self:
             if not line.order_line_ids:
                 return
-            line.prod_lot_ids = self.mapped(
-                'order_line_ids.procurement_ids.move_ids').filtered(
-                lambda move: move.state == 'done').lot_ids
+            move_ids = self.mapped(
+                'order_line_ids.procurement_ids.move_ids')
+            prod_lot_ids = move_ids.filtered(
+                lambda move: move.state == 'done').mapped('lot_ids')
+            line.prod_lot_ids = prod_lot_ids
 
     @api.multi
     def _compute_line_lots(self):
--- a/account_invoice_production_lot/tests/test_invoice_production_lot.py
+++ b/account_invoice_production_lot/tests/test_invoice_production_lot.py
@@ -2,6 +2,7 @@
 # Copyright 2011 Domsense s.r.l. <http://www.domsense.com>
 # Copyright 2013 Lorenzo Battistini <lorenzo.battistini@agilebg.com>
 # Copyright 2017 Vicent Cubells <vicent.cubells@tecnativa.com>
+# Copyright 2018 Alex Comba <alex.comba@agilebg.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo.tests import common
@@ -17,6 +18,7 @@ class TestProdLot(common.SavepointCase):
         cls.product = cls.env['product.product'].create({
             'name': 'Product Test',
             'type': 'product',
+            'tracking': 'lot',
         })
         cls.sale = cls.env['sale.order'].create({
             'partner_id': cls.partner.id,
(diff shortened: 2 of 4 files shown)


account_invoice_production_lot/models/account_invoice.py
```diff
@@ -36,7 +36,8 @@ def _compute_prod_lots(self):
             if not line.order_line_ids:
                 return
             line.prod_lot_ids = self.mapped(
-                'order_line_ids.procurement_ids.move_ids.lot_ids')
+                'order_line_ids.procurement_ids.move_ids').filtered(
+                lambda move: move.state == 'done').lot_ids
```
SimoRubi: Note that if there are multiple done moves, this gets an error: you should retrieve `lot_ids` using `mapped`.
Moreover, maybe it is more readable if you split all this in a few lines.

account_invoice_production_lot/tests/test_invoice_production_lot.py
```diff
@@ -78,3 +90,73 @@ def test_invoice_product_lot(self):
...
+        backorder_wiz = self.env['stock.backorder.confirmation'].browse(
+            [backorder_wiz_id])
+        backorder_wiz.process()
+        # create invoice
+        inv_id = self.sale.action_invoice_create()
+        invoice = self.env['account.invoice'].browse(inv_id)
+        self.assertEqual(len(invoice.invoice_line_ids), 1)
+        line = invoice.invoice_line_ids
+        # We must have only one lot
+        self.assertEqual(len(line.prod_lot_ids.ids), 1)
```
SimoRubi, line 125: Why don't you assert that this is the lot you want using its `id`?

account_invoice_production_lot/tests/test_invoice_production_lot.py
```diff
@@ -78,3 +90,73 @@ def test_invoice_product_lot(self):
...
+        self.assertIn('Lot 1', line.lot_formatted_note)
```
SimoRubi, line 127: Looking at the usecase you described, here you should also assert that `'Lot 2'` is not in `line.lot_formatted_note`.
Anyway you can just skip this assertion if in the line above you check using the lot's `id`.

account_invoice_production_lot/tests/test_invoice_production_lot.py
```diff
@@ -78,3 +90,73 @@ def test_invoice_product_lot(self):
...
+        backorder_wiz = self.env['stock.backorder.confirmation'].browse(
+            [backorder_wiz_id])
+        backorder_wiz.process()
+        # create invoice
+        inv_id = self.sale.action_invoice_create()
+        invoice = self.env['account.invoice'].browse(inv_id)
+        self.assertEqual(len(invoice.invoice_line_ids), 1)
+        line = invoice.invoice_line_ids
+        # We must have two lots
+        self.assertEqual(len(line.prod_lot_ids.ids), 2)
```
SimoRubi, line 161: Same remarks as previous test's

SimoRubi (changes requested):
Code review

tafaRU:
@SimoRubi, done in https://github.com/OCA/account-invoice-reporting/pull/84/commits/26d051d3487c1c96a91c49e25ceab553a754ab2b Thanks!

SimoRubi:
@tafaRU what about the tests?
Why don't you check the lots using their `id` instead of `self.assertIn('Lot 1', line.lot_formatted_note)`?
Wouldn't that be more straightforward?

tafaRU:
@SimoRubi, I checked both conditions, isn'it?
I'd prefer kept that one related to `lot_formatted_note` in order to also check the string used in the invoice report.

SimoRubi:
Ok I thought you had experienced some trouble in comparing the `id`s instead of comparing the strings, but if it's just your choice then ok.
Anyway tested this in runbot and it works!
Thanks for your contribution

SimoRubi (approved):
Tested in Runbot too

tafaRU:
> The remarks of my previous still stand

@SimoRubi, It seems to me they have been already addressed in https://github.com/OCA/account-invoice-reporting/commit/26d051d3487c1c96a91c49e25ceab553a754ab2b

**test_01_sale_stock_delivery_partial_invoice_product_lot**
```
self.assertEqual(line.prod_lot_ids.id, self.lot1.id)
```

**test_02_sale_stock_delivery_partial_invoice_product_lot**
```
self.assertIn(self.lot1.id, line.prod_lot_ids.ids)
self.assertIn(self.lot2.id, line.prod_lot_ids.ids)
```
Isnt'it?

tafaRU:
Superseded by https://github.com/OCA/account-invoice-reporting/pull/85

tafaRU:
@SimoRubi, thank you for your review.
Could you also please review https://github.com/OCA/account-invoice-reporting/pull/85?

PR #88, opened 2018-10-19
[12.0] transifex --> weblate

eLBati:


tafaRU (approved):

pedrobaeza (approved):
Well, in fact we keep a compatibility layer that makes `TRANSIFEX` stage to act like `MAKEPOT` one, so it's working right now, but with this we clean the file and reduce one build.

PR #89, opened 2018-10-21
[UPD] Update id.po

rsanhasan:
- Add id.po
- Update id.po


pedrobaeza:
Please use Weblate instead: https://odoo-community.org/page/translate

PR #90, opened 2018-10-29
 [MIG] account_reporting_weight: Migration to 11.0

chienandalu:
This module adds the measure "Weight" in the invoices analysis view. This is
caught from 2 possible sources:

* If the UoM of the product is one of the category "Weight", the value is taken
  from the ordered quantity.
* If the UoM of the product is another, then the weight is taken from the
  weight field of the product multiply by the ordered quantity.

cc @Tecnativa


pedrobaeza:
Shouldn't we use only the weight here and `product_weight_through_uom`?

chienandalu:
The weight measure is not available in this report. Anyway, we can make it dependant on `product_weight_through_uom` and simplify the query

pedrobaeza:
Yes, what I'm proposing is to simplify query, but no need to depend on the other module: if you have it, better for data integrity, but no real need.

chienandalu:
Done

pedrobaeza (approved):

cubells (approved):
:+1: 

LGTM

Code review and test

cristinamartinrod (approved):
Works fine 👍🏻

PR #93, opened 2018-11-30
[10.0][FIX] account_invoice_production_lot: handle return picking.

tafaRU:
These changes are intended to cover the following use case scenario.

Steps to reproduce:

* create a new product 'Product Test': stockable, Invoicing Policy as Delivered quantities, and tracking by lots
* update its quantity to 4 with new lot 'Lot1'
* update its quantity to 4 with new lot 'Lot2'
* sale 4 of 'Product Test', confirm the SO
* on the delivery, split the lots and set To Do quantities as 4 for Lot1
* validate the delivery
* return all quanties related to 'Lot 1'
* duplicate the original delivery and use 'Lot 2' instead of 'Lot 1'
* validate the new delivery

Current behavior:

Expected behavior:

the invoice report only contains info related to the lot involved in the new delivery, which is `Lot 2`.


account_invoice_production_lot/models/account_invoice.py
```diff
@@ -31,10 +31,21 @@ class AccountInvoiceLine(models.Model):
         compute='_compute_line_lots',
     )
 
+    @api.multi
+    def _get_returned_moves(self):
+        self.ensure_one()
+        procurement_ids = self.move_line_ids.mapped('procurement_id').ids
+        moves = self.env['stock.move'].search(
+            [('procurement_id', 'in', procurement_ids)]).filtered(
+            lambda m: m.state == ('done'))
```
SimoRubi: Why don't you insert this in the domain?
One more question: here you are getting all the done moves linked to the procurement, how can you be sure that they are moves for returned products?

SimoRubi (commented):
Code review

account_invoice_production_lot/models/account_invoice.py
tafaRU: > Why don't you insert this in the domain?

Done in https://github.com/OCA/account-invoice-reporting/pull/93/commits/d8e31422ebd29eb9632a5c957c93ec3020eecefa

> how can you be sure that they are moves for returned products?

Checking if `origin_returned_move_id` is filled (see https://github.com/odoo/odoo/blob/12.0/addons/stock/wizard/stock_picking_return.py#L87).
And by mapping as follows:

`moves.mapped('origin_returned_move_id')`

I get all the moves involved in the return.

SimoRubi: oh ok I didn't see the last line in the `return`, thanks

SimoRubi (approved):
Code review ok for me. Thanks!

eLBati (approved):
LGTM

account_invoice_production_lot/models/account_invoice.py
```diff
@@ -51,6 +62,7 @@ def _compute_line_lots(self):
                 note += u'</ul>'
                 line.lot_formatted_note = note
 
+    @api.multi
```
pedrobaeza, line 65: There's no need to put the decorator: it's the default one when there's nothing.

pedrobaeza (changes requested):
You should consider recursively all the possible returns, and discriminate through the destination location, as you can return a return, but also return a return of a return...

tafaRU:
> You should consider recursively all the possible returns, and discriminate through the destination location, as you can return a return, but also return a return of a return...

@pedrobaeza, could you please elaborate a bit more?

EDIT: also please note that this PR is only intended to cover the use case scenario described. Maybe the commit message header is misleading!

account_invoice_production_lot/models/account_invoice.py
tafaRU: ok, good to know it, but according to the Zen of Python: `Explicit is better than implicit.` :wink:

pedrobaeza: But this is not Python, but Odoo ORM :wink:

pedrobaeza:
I think I have explained it. Let me rephrase it again: doing a return of a return will hide some valid lots in the invoice.

tafaRU:
Superseeded by https://github.com/OCA/account-invoice-reporting/pull/99

PR #98, opened 2019-01-21
[ADD][10.0] new module company_invoice_note

acsonefho:
This module add a note field on the company who'll be printed on every invoice reports.
This can be printed on different places:

- Before the invoice comment;
- Instead of the invoice comment;
- At the end of the document (after fiscal position note).


pedrobaeza:
The same said with `base_comment_template`, but for `account_invoice_comment_template`

acsonefho:
@pedrobaeza The goal is not exactly the same. Here we want to add a comment/note. We don't need a model to save some possible comments because we only have 1 value (per company). It's more like general conditions to add on the report.

pedrobaeza:
But you can achieve what you want with the other module, having defaults per company. For me this is totally overlapping, so :-1:

PR #99, opened 2019-01-22
[FIX] account_invoice_production_lot: handle return picking.

tafaRU:
These changes are intended to cover the following use case scenario.

Steps to reproduce:

* create a new product 'Product Test': stockable, Invoicing Policy as Delivered quantities, and tracking by lots
* update its quantity to 4 with new lot 'Lot1'
* update its quantity to 4 with new lot 'Lot2'
* sale 4 of 'Product Test', confirm the SO
* on the delivery, split the lots and set To Do quantities as 4 for Lot1
* validate the delivery
* return all quanties related to 'Lot 1'
* return of a return and use 'Lot 2' instead of 'Lot 1'
* validate the new delivery

Current behavior:

Expected behavior:

the invoice report only contains info related to the lot involved in the new delivery, which is 'Lot 2'.


account_invoice_production_lot/models/account_invoice.py
```diff
@@ -31,10 +31,18 @@ class AccountInvoiceLine(models.Model):
         compute='_compute_line_lots',
     )
 
+    @api.multi
+    def _get_location_id(self):
+        self.ensure_one()
+        order = self.sale_line_ids[0].order_id
+        return order.partner_shipping_id.property_stock_customer
```
pedrobaeza: I don't think this is the proper location to use for all cases. Why not using the location type = 'internal'?

tafaRU: I had a try locally with:

```
    @api.multi
    def _compute_prod_lots(self):
        for line in self:
            quant_ids = line.move_line_ids.mapped('quant_ids').filtered(
                lambda x: x.location_id.usage == 'internal')
            line.prod_lot_ids = quant_ids.mapped('lot_id')
```
but tests are failing:
`2019-01-22 15:07:30,175 20785 ERROR test-account-invoice-reporting-93 odoo.modules.module: Module account_invoice_production_lot: 3 failures, 0 errors
`

I take the same approach of https://github.com/odoo/odoo/blob/10.0/addons/sale_stock/models/sale_order.py#L167

account_invoice_production_lot/models/account_invoice.py
```diff
@@ -31,10 +31,18 @@ class AccountInvoiceLine(models.Model):
         compute='_compute_line_lots',
     )
 
+    @api.multi
+    def _get_location_id(self):
```
SimoRubi: Maybe this name is too generic, it might interfere with some other `_get_location_id` methods: I would prefer a more specific name.

account_invoice_production_lot/models/account_invoice.py
```diff
@@ -31,10 +31,18 @@ class AccountInvoiceLine(models.Model):
...
+        self.ensure_one()
+        order = self.sale_line_ids[0].order_id
+        return order.partner_shipping_id.property_stock_customer
+
     @api.multi
     def _compute_prod_lots(self):
         for line in self:
-            line.prod_lot_ids = line.move_line_ids.mapped('lot_ids')
+            quant_ids = line.move_line_ids.mapped('quant_ids').filtered(
+                lambda x: x.location_id == line._get_location_id())
```
SimoRubi: Get the location once, outside the `filtered` method: otherwise here it might get uselessly called many times (once for each quant) for each line.

SimoRubi (changes requested):
Thanks!
Code review and tried on runbot

account_invoice_production_lot/models/account_invoice.py
tafaRU: Done. See: https://github.com/OCA/account-invoice-reporting/compare/808b4d36e3264296f1a08589c7d22e9d2cc6e5ca..f37f03e8af1380685e31ccc7b91d6d7d609fe259

account_invoice_production_lot/models/account_invoice.py
tafaRU: Done. See: https://github.com/OCA/account-invoice-reporting/compare/808b4d36e3264296f1a08589c7d22e9d2cc6e5ca..f37f03e8af1380685e31ccc7b91d6d7d609fe259

account_invoice_production_lot/models/account_invoice.py
```diff
@@ -31,10 +31,19 @@ class AccountInvoiceLine(models.Model):
...
+        self.ensure_one()
+        order = self.sale_line_ids[0].order_id
+        return order.partner_shipping_id.property_stock_customer
+
     @api.multi
     def _compute_prod_lots(self):
         for line in self:
-            line.prod_lot_ids = line.move_line_ids.mapped('lot_ids')
+            location_id = line._get_stock_customer_location_id()
+            quant_ids = line.move_line_ids.mapped('quant_ids').filtered(
```
SimoRubi, line 37: You should probably identify the delivered lots in a similar way to [how the `sale.order.line` computes the delivered qty](https://github.com/odoo/odoo/blob/0b414cb8145381cd8a3946aae9cec1601b111238/addons/sale_stock/models/sale_order.py#L176-L189).

Best way in my opinion is to start from `line.move_line_ids` (as you are already doing here).

SimoRubi (changes requested):
A little concern about how to compute delivered lots

account_invoice_production_lot/models/account_invoice.py
tafaRU: @SimoRubi did you mean https://github.com/tafaRU/account-invoice-reporting/commit/4697ce67ee33db1cb72844e7dd4ebc8932621101?
Tests are all passed :heavy_check_mark:

SimoRubi: yes @tafaRU that is the base idea, but the changes of https://github.com/tafaRU/account-invoice-reporting/commit/4697ce67ee33db1cb72844e7dd4ebc8932621101 are not included in this PR

tafaRU: Ok thanks.

> but the changes of [tafaRU@4697ce6](https://github.com/tafaRU/account-invoice-reporting/commit/4697ce67ee33db1cb72844e7dd4ebc8932621101) are not included in this PR

Yep, I was waiting for your confirmation!

tafaRU: @SimoRubi just force pushed. Please update your review.

SimoRubi (commented):
Same concern as earlier

account_invoice_production_lot/models/account_invoice.py
```diff
@@ -34,7 +34,9 @@ class AccountInvoiceLine(models.Model):
     @api.multi
     def _compute_prod_lots(self):
         for line in self:
-            line.prod_lot_ids = line.move_line_ids.mapped('lot_ids')
+            quant_ids = line.move_line_ids.mapped('quant_ids').filtered(
+                lambda x: x.location_id.usage == 'customer')
```
SimoRubi, line 38: @tafaRU why didn't you keep the other filters on the moves too?
```
not move.origin_returned_move_id or (move.origin_returned_move_id and move.to_refund_so)
```
source (from the same method linked earlier): https://github.com/odoo/odoo/blob/0b414cb8145381cd8a3946aae9cec1601b111238/addons/sale_stock/models/sale_order.py#L185

SimoRubi (commented):
comment

account_invoice_production_lot/models/account_invoice.py
SimoRubi: @tafaRU I have created https://github.com/tafaRU/account-invoice-reporting/pull/1.
The filters that I suggested in https://github.com/OCA/account-invoice-reporting/pull/99#discussion_r255967022 aren't completely feasible with this module's purpose because, for what I read from the README and the tests, it doesn't care about:

- refunds: field `to_refund_so` in the return wizard is never mentioned
- the `quantity` returned: if a lot is returned, it counts as returned regardless it was completely returned or not.

That's why I only took a few lines from the core implementations, and tried to manage returned moves my way.
I added comments too, hope it's clear enough.

eLBati: > - refunds: field to_refund_so in the return wizard is never mentioned
> - the quantity returned: if a lot is returned, it counts as returned regardless it was completely returned or not


@SimoRubi the implementation based on quants available in customer location should not be affected by this problems

SimoRubi: @eLBati in https://github.com/tafaRU/account-invoice-reporting/pull/1 I tried to implement a solution to what was concerning me in https://github.com/OCA/account-invoice-reporting/pull/99#discussion_r255967022: I tried to mimic the behavior of Odoo core in order to have consistent behavior.

If Odoo core counts the delivered qty [this way](https://github.com/odoo/odoo/blob/0b414cb8145381cd8a3946aae9cec1601b111238/addons/sale_stock/models/sale_order.py#L176-L189) maybe we should do so too and there are some implications that we don't know in looking only to quants.

eLBati: Here we need the delivered lots, not the quantities

SimoRubi:
Right, but the computation should be similar because delivered lots and
delivered quantities have both been delivered in the same moves.
I have already stated the differences a few posts above and in the comments
of the code

eLBati:
I can't see why @tafaRU should move again to another (a more complex one) implementation of `_compute_prod_lots` when no possible reasons of failure are found.

From what I can see at https://github.com/tafaRU/account-invoice-reporting/pull/1/  I think it's highly risky to base the computation on moves instead of quants, as moves can represent complicated movements in picking elaboration, hard to predict.

Anyway please provide the failing test case.

SimoRubi:
I don't have a failing test case: I just think, for the reasons explained
above, that it might be a better implementation because it is more similar
to what Odoo core does for a similar purpose (computing delivered quantity).

PR #101, opened 2019-02-20
[11.0][FIX] account_invoice_report_grouped_by_picking: Error division by zero when quantity is zero

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/account_invoice_report_grouped_by_picking/views/report_invoice.xml
+++ b/account_invoice_report_grouped_by_picking/views/report_invoice.xml
@@ -40,7 +40,7 @@
     <xpath expr="//td/span[@t-field='l.price_subtotal']" position="before">
         <t t-if="picking" id="picking_subtotal" groups="!sale.group_show_price_total">
             <!-- Compute subtotal for that picking with discounts -->
-            <t t-set="line_picking_subtotal" t-value="l.quantity and lines_group['quantity'] * (l.price_subtotal / l.quantity or 1.0) or 0.0"/>
+            <t t-set="line_picking_subtotal" t-value="l.quantity and lines_group['quantity'] * (l.price_subtotal / l.quantity) or 0.0"/>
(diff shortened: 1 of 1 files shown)


account_invoice_report_grouped_by_picking/views/report_invoice.xml
```diff
@@ -40,7 +40,7 @@
     <xpath expr="//td/span[@t-field='l.price_subtotal']" position="before">
         <t t-if="picking" id="picking_subtotal" groups="!sale.group_show_price_total">
             <!-- Compute subtotal for that picking with discounts -->
-            <t t-set="line_picking_subtotal" t-value="lines_group['quantity'] * (l.price_subtotal / l.quantity)"/>
+            <t t-set="line_picking_subtotal" t-value="l.quantity and lines_group['quantity'] * (l.price_subtotal / l.quantity or 1.0) or 0.0"/>
```
pedrobaeza: The `or 1.0` part is not correct and can even lead to errors in the computation not being between parenthesis:
```
>>> (0.0 / 1.0 or 1.0)
1.0
```
Please remove it or put it between brackets (but it's not needed as you already added the previous condition):
```xml
(l.price_subtotal / (l.quantity or 1.0))
```

pedrobaeza (changes requested):

account_invoice_report_grouped_by_picking/views/report_invoice.xml
carlosdauden:
```suggestion
            <t t-set="line_picking_subtotal" t-value="l.quantity and lines_group['quantity'] * (l.price_subtotal / l.quantity) or 0.0"/>
```

carlosdauden (approved):

pedrobaeza (approved):

PR #104, opened 2019-03-29
[12.0][MIG] account_invoice_production_lot

AdriaGForgeFlow:
Standard migration of account_invoice_production_lot module to v12


pedrobaeza:
It's interesting to review the related PR #99.

FYI, Odoo v13 will include a mechanism for putting already this information in invoice, so hopefully this module won't be needed.

pedrobaeza:
Are you going to fix Travis and runbot?

account_invoice_production_lot/models/account_invoice.py
```diff
@@ -0,0 +1,49 @@
...
+                            _('S/N'), sml.lot_id.name,
+                        ))
+                    else:
+                        lot_strings.append('<li>%s %s (%s)</li>' % (
+                            _('Lot'), sml.lot_id.name, sml.qty_done,
+                        ))
+            if lot_strings:
+                note += ' '.join(lot_strings)
+                note += '</ul>'
+                line.lot_formatted_note = note
```
sergio-teruel: Why not group by lots...? I attach a picture. By this way we cosume a lot vertical space
And is posible do this directly in report?

Selección_299

pedrobaeza:
@jbeficent can you attend this?

account_invoice_production_lot/models/account_invoice.py
AdriaGForgeFlow: Lots should be grouped correctly now

sergio-teruel:
Please fix travis.. 
I think that field lot_formatted_note is not very inheritable..
Mybe you could return a list of dictionaries with lot and units and do render in a report view.
¿What do you think?

sergio-teruel:
PR superseed in https://github.com/OCA/account-invoice-reporting/pull/123

PR #119, opened 2019-10-16
13.0 mig base comment template

emagdalenaC2i:


emagdalenaC2i:
@pedrobaeza I don't know why travis is red. I did the same than in other migration PR
Some clue?

pedrobaeza:
The error is "Check that executables have shebangs", which means that you have marked any file as executable.

emagdalenaC2i:
But i didn't do that

pedrobaeza:
Ok, but it should be fixed any way

PR #126, opened 2019-12-13
[12.0][IMP] - Use Jinja2 template to customize the payment mode note

sbejaoui:
allow users to customize the payment mode note using jinja2 templates


gva-acsone (approved):
Functional test LGTM

pedrobaeza:
Isn't this interesting to be in `account_payment_mode` + `account_payment_partner`?

PR #127, opened 2019-12-13
[11.0][IMP] account_invoice_report_grouped_by_picking: Group service lines also with the related SO picking

pedrobaeza:
Service lines that are not put on the pickings, remain on top of the invoice report due to this, but it has more sense to have them grouped in the same picking block than the rest of their sales order stockable lines.
    
Tests and README adapted

@Tecnativa TT21060


carlosdauden (approved):

sergio-teruel (approved):

pedrobaeza:
Cherry-picked for v12 in 0f452971184f8de82cb8ad5b8d9fb12238bbc904

PR #130, opened 2020-01-24
[IMP] account_invoice_production_lot: serials rendering

chienandalu:
When the product traceability is by unique serial numbers, each unit on
the invoice line is going to be shown on a separate row, wich can result
in quite long reports. So in these cases is preferible to show them
altogether separated by commas.

cc @Tecnativa TT21255


pedrobaeza (approved):

Tardo (approved):

pedrobaeza:
Please do the same for v12

/ocabot merge minor

PR #132, opened 2020-01-29
[IMP] account_invoice_report_due_list: allways print the same way

chienandalu:
- It's annoying to have the due date rendered in different positions
depending on if it's a single due or there are several. With this
change, we render it always the same way.

cc @Tecnativa TT21255

PR #133, opened 2020-01-29
[IMP] account_invoice_report_due_list: allways print the same way

chienandalu:
- It's annoying to have the due date rendered in different positions
depending on if it's a single due or there are several. With this
change, we render it always the same way.

cc @Tecnativa TT21255

FW Port from https://github.com/OCA/account-invoice-reporting/pull/132


pedrobaeza (approved):
Fast-tracking this as merged in 11.0

PR #134, opened 2020-02-18
[FIX] account_invoice_report_due_list: empty table

chienandalu:
When no due dates are available we get an empty table. This patch fixes
it so the due list table is only visible when due dates are computed.

cc @Tecnativa TT21247


pedrobaeza (approved):
Fast-tracking this and please forward-port it

PR #135, opened 2020-02-18
[FIX] account_invoice_report_due_list: empty table

chienandalu:
When no due dates are available we get an empty table. This patch fixes
it so the due list table is only visible when due dates are computed.

@Tecnativa TT21247

FW Port of https://github.com/OCA/account-invoice-reporting/pull/134


pedrobaeza:
Fast-tracking as well

/ocabot merge minor

Is this in v13 already?

PR #137, opened 2020-03-10
[13.0][MIG] account_invoice_report_grouped_by_picking : Migration to 13.0

manuelcalerosolis:
This is the migration of module mig_account_invoice_report_grouped_by_picking to 13.0

@Tecnativa TT21605

Depends on:
[WIP][MIG][13.0] stock_picking_invoice_link: Migration to v13.0 OCA/stock-logistics-workflow#/622


pedrobaeza:
Why this depends on project_timesheet_time_control? Shouldn't it be `stock_picking_invoice_link`?

manuelcalerosolis:
Sorry my mistake. The dependency must be stock_picking_invoice_link in its version 13, try to clarify with @sergio-teruel  how this task goes. For that reason it is still marked as WIP.

pedrobaeza:
Can you migrate that module to v13 including the ongoing PR for v12?

manuelcalerosolis:
One issue to clarify is that it already appears in the repository as completely migrated in 12.0

https://github.com/OCA/stock-logistics-workflow/tree/12.0/stock_picking_invoice_link

pedrobaeza:
It's migrated, but there's a problem that is solved in that PR, and that must be fixed before even trying to migrate it to v13.

Tardo:
Check Travis

sergio-teruel:
@carlosdauden @pedrobaeza Ready to be merged!!

account_invoice_report_grouped_by_picking/security/portal_security.xml
```diff
@@ -1,16 +1,16 @@
...
-        <field name="domain_force">['|', ('partner_id.id', 'child_of', user.partner_id.commercial_partner_id.id), ('sale_id.partner_id.id', 'child_of', user.partner_id.commercial_partner_id.id)]</field>
-        <field name="groups" eval="[(4, ref('base.group_portal'))]"/>
-        <field name="perm_read" eval="True"/>
-        <field name="perm_write" eval="False"/>
-        <field name="perm_create" eval="False"/>
-        <field name="perm_unlink" eval="False"/>
+        <field name="model_id" ref="stock.model_stock_picking" />
+        <field
+            name="domain_force"
+        >['|', ('partner_id.id', 'child_of', user.partner_id.commercial_partner_id.id), ('sale_id.partner_id.id', 'child_of', user.partner_id.commercial_partner_id.id)]</field>
```
pedrobaeza: I don't think this is the best approach for avoiding the access problem, as what happened to a public user with a token for accessing the invoice?

I think the solution is to put sudo where needed in the QWeb report. Note that this problem is also in v12, but I have detected it now.

pedrobaeza (changes requested):

account_invoice_report_grouped_by_picking/security/portal_security.xml
sergio-teruel: OH! yes!
https://github.com/odoo/odoo/blob/60704b3151f66b4daf87c147afcd231ae0713ab3/addons/account/controllers/portal.py#L82
The invoices are rendered as sudo, so we can remove this rule

sergio-teruel: In v12 we can remove it too

pedrobaeza (approved):

PR #142, opened 2020-04-21
[FIX] account_invoice_report_due_list: avoid breaking due list block

chienandalu:
When the due list table falls in a page break it's not properly
rendered. It's more convinient to keep its integrity anyway.

@Tecnativa TT23478


pedrobaeza (approved):

bodedra (approved):
LGTM :+1:

carlosdauden (approved):

PR #151, opened 2020-07-24
[12.0][ADD] account_invoice_report_grouped_by_picking_sale_mrp: new module

ernestotejeda:
I made a hook method in 'account.invoice' inside account_invoice_report_grouped_by_picking module to be uses in this module.

cc @Tecnativa TT24818


account_invoice_report_grouped_by_picking_sale_mrp/README.rst
```diff
@@ -0,0 +1,99 @@
...
+
+#. Go to *Manufacturing -> Master Data -> Bills of Materials* and create a
+   new BoM selecting the product, 'Kit' in 'BoM Type' field and filling
+   the components list.
+#. Go to *Sales -> Orders -> Quotations* and create a new quotation
+   with the 'Kit' previously created and qty 1.
+#. Confirm the sale order and you will see one delivery (accessible
+   via smart-button) with the components of the kit.
+#. Validate the delivery.
+#. Go back to the sale order and increase the ordered qty by 1, save, and
```
pedrobaeza: There's no need for this second picking extra step for the basic explanation of the module.

account_invoice_report_grouped_by_picking_sale_mrp/README.rst
```diff
@@ -0,0 +1,99 @@
...
+   you will see another delivery linked to the sale order with the
+   components of the kit.
+#. Validate this second delivery.
+#. Go back to the sale order and create an invoice by clicking on
+   Create Invoice button in the status bar.
+#. Print the invoice and the generated report will have 2 group:
+    * One group with the information of the sale order and the first delivery
```
pedrobaeza: Incorrect indentation (one space less), and you have to leave one empty blank line between both levels.

account_invoice_report_grouped_by_picking_sale_mrp/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2020 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Account Invoice Grouped by Picking - Sale MRP",
+    "summary": "Print invoice lines grouped by picking, but taking "
+               "into account MRP kit sales",
```
pedrobaeza:
```suggestion
    "summary": "Take into account BoM kits in invoice report grouped by picking",
```

account_invoice_report_grouped_by_picking_sale_mrp/models/mrp_bom.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class MrpBom(models.Model):
+    _inherit = 'mrp.bom'
+
+    components_total_qty = fields.Float(
+        compute="_compute_components_total_qty",
+    )
+
+    def _compute_components_total_qty(self):
```
pedrobaeza: Don't create a computed field if it's going to be used only in one method. This have an overhead at ORM level.

account_invoice_report_grouped_by_picking_sale_mrp/models/mrp_bom.py
```diff
@@ -0,0 +1,23 @@
...
+        for record in self:
+            qty = 0
+            for line in record.bom_line_ids:
+                line_qty = line.product_qty
+                line_bom = self._bom_find(product=line.product_id)
```
pedrobaeza: Isn't this code anywhere in MRP for avoiding to duplicate it?

pedrobaeza (changes requested):
Almost good. Some details in inline comments.

account_invoice_report_grouped_by_picking_sale_mrp/models/mrp_bom.py
ernestotejeda: Well, in the mrp module in odoo core I didn't see any method that would do that.

ernestotejeda:
@pedrobaeza changes done

account_invoice_report_grouped_by_picking_sale_mrp/models/mrp_bom.py
pedrobaeza: What about using `mrp.bom`.`explode` method for fetching data and obtain final quantity from there?

ernestotejeda: Yes, I just checked it in detail and it is true, I can use it. Thank you @pedrobaeza

ernestotejeda:
@pedrobaeza  Changes done

pedrobaeza (approved):

carlosdauden (approved):

PR #159, opened 2020-11-23
[12.0] ADD account_invoice_package

eLBati:
This module shows, for each (customer) invoice line, the delivered packages
that will be also displayed on the invoice report with their corresponding
delivered quantities.


mrcast (approved):
I tested on a local database and it's ok

PR #162, opened 2021-01-01
[13.0][IMP] account_invoice_report_grouped_by_picking: No display repeated lines for refund invoices without no returned quantities

sergio-teruel:
cc @Tecnativa TT27550
ping @carlosdauden 

Before this PR:
image

After this PR:
image

### DIFF
--- a/account_invoice_report_grouped_by_picking/models/account_move.py
+++ b/account_invoice_report_grouped_by_picking/models/account_move.py
@@ -29,6 +29,7 @@ class AccountMove(models.Model):
         so_dict = {x.sale_id: x for x in self.picking_ids if x.sale_id}
         # Now group by picking by direct link or via same SO as picking's one
         for line in self.invoice_line_ids:
+            has_returned_qty = False
             remaining_qty = line.quantity
             for move in line.move_line_ids:
                 key = (move.picking_id, line)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
But should we keep the original amount in the line in these cases? Some customers will expect so, and other will expect to have 0 quantity.

sergio-teruel:
@pedrobaeza is for v13.0
I do not understand your approach

pedrobaeza:
Having this situation:

- Delivered 10 units
- Refund that 10 units via accounting, not via return picking.

What do you want to show in the invoice? 10 or 0 units?

sergio-teruel:
Se can discuss about your query, but this PR does not change these behavior.
I do not know if user expects 0 or 10 units but O am sure that user not expects a line with 20 units and another one with -10 units...

pedrobaeza:
OK, let's try to fix that one. Can you add a regression test for the use case?

sergio-teruel:
@pedrobaeza Tests added!!

sergio-teruel:
I am seeing that some lines has not been covered by tests ... 
Plase wait to improve the tests for merge it...

chusamo (approved):
Tested in Local LGTM

carlosdauden (approved):

PR #164, opened 2021-01-08
[V13] Add account invoice bank details

vrenaville:
* Purpose
If you need to display bank information on invoice report, it also select the account linked to to right currency if present.

### DIFF
--- a/account_invoice_bank_details/__init__.py
+++ b/account_invoice_bank_details/__init__.py
@@ -1,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import models
--- a/account_invoice_bank_details/__manifest__.py
+++ b/account_invoice_bank_details/__manifest__.py
@@ -1,0 +1,16 @@
+{
+    "name": "Invoice bank account details",
+    "summary": "Select bank account base on currency + print bank details on report"
+    "reports and customer portal",
+    "version": "13.0.1.0.0",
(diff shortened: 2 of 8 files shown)


pedrobaeza:
This is already done with `account_payment_partner` module. Do you know it?

santostelmo (approved):

fclementic2c:
@pedrobaeza Hello, as far as I remember account_payment_partner does not display the bank details on the invoice report according to the currency... am I wrong ?

pedrobaeza:
Yes, it does, but not according the currency, but the payment mode:

https://github.com/OCA/bank-payment/blob/7c72a0b68bac93588e11f97d29612774505a636a/account_payment_partner/views/report_invoice.xml#L12

For me it's not a good design to base something on the currency used. You have payment modes for reflecting that IMO.

fclementic2c:
@pedrobaeza ok thanks. But why do you say that it is a bad design to base it on currency ? Because this is what customers are always asking for and it sounds rational to me.

pedrobaeza:
I say it because it's not very flexible, as you are coupling bank selection to a very short list of options. Payment modes are more flexible.

fclementic2c:
ok. thanks for the feedback. I guess this a problem of volumetry. When create a lot of invoices you may want to set up rules to select the correct payment mode. It could be currency, country of destination, type of products... maybe it deserve a module on top of account_payment_partner.

PR #168, opened 2021-01-18
[14.0][FIX] account_invoice_report_grouped_by_picking: invoice line without product

miguel-S73:
Odoo returns an error when there is a note or section line on the invoice.
In older versions like odoo 12 it also happens. Tested in runbot.

### DIFF
--- a/account_invoice_report_grouped_by_picking/models/account_move.py
+++ b/account_invoice_report_grouped_by_picking/models/account_move.py
@@ -49,7 +49,8 @@ class AccountMove(models.Model):
                         picking_dict[key] += qty
                         remaining_qty -= qty
             if not float_is_zero(
-                remaining_qty, precision_rounding=line.product_id.uom_id.rounding
+                remaining_qty,
+                precision_rounding=line.product_id.uom_id.rounding or 0.01,
             ):
                 lines_dict[line] = remaining_qty
         no_picking = [


pedrobaeza:
I don't get the case where you have a note/section and quantity is not zero. Anyway, the proper fix is not to check about `product_id` (you can have invoice lines without product_id), but line type.

miguel-S73:
> I don't get the case where you have a note/section and quantity is not zero. Anyway, the proper fix is not to check about `product_id` (you can have invoice lines without product_id), but line type.

Error occurs because there isn't `product_id`. 
image

account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -48,7 +48,7 @@ def lines_grouped_by_picking(self):
                         qty = so_line.product_uom_qty
                         picking_dict[key] += qty
                         remaining_qty -= qty
-            if not float_is_zero(
+            if line.product_id and not float_is_zero(
                 remaining_qty, precision_rounding=line.product_id.uom_id.rounding
```
pedrobaeza:
```suggestion
            if not float_is_zero(
                remaining_qty, precision_rounding=line.product_id.uom_id.rounding or 0.01
```

pedrobaeza (commented):
Then this is the proper fix

account_invoice_report_grouped_by_picking/models/account_move.py
miguel-S73: `line.product_id.uom_id.rounding` this already returns 0. The function checks that the value is greater than 0 and since it is not, the error occurs.

pedrobaeza: Rectified my suggestion then

miguel-S73: Change applied

account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -48,8 +48,10 @@ def lines_grouped_by_picking(self):
                         qty = so_line.product_uom_qty
                         picking_dict[key] += qty
                         remaining_qty -= qty
+
```
pedrobaeza: Why adding a useless extra empty line? PEP8 discourages this, and I personally hate it.

pedrobaeza (approved):

account_invoice_report_grouped_by_picking/models/account_move.py
miguel-S73: Fixed

ioans73 (approved):

pedrobaeza:
This should be backported if applicable to lower versions.

PR #170, opened 2021-01-22
13.0 [FIX] account_invoice_report_due_list, error on empty date_maturity

manuelcalerosolis:
When the date_maturity field does not contain a date, the module throws an error, trying to convert the field to a string with strftime (). 

In addition, this error does not allow inserting accounting entries in the accounting.

### DIFF
--- a/account_invoice_report_due_list/models/account_move.py
+++ b/account_invoice_report_due_list/models/account_move.py
@@ -30,11 +30,12 @@ class AccountMove(models.Model):
         due_list = []
         if self.type in ["in_invoice", "out_refund"]:
             due_move_line_ids = self.line_ids.filtered(
-                lambda ml: ml.account_id.internal_type == "payable"
+                lambda ml: ml.account_id.internal_type == "payable" and ml.date_maturity
             )
         else:
             due_move_line_ids = self.line_ids.filtered(
                 lambda ml: ml.account_id.internal_type == "receivable"
+                and ml.date_maturity
             )
         if self.currency_id != self.company_id.currency_id:
             due_list = [


pedrobaeza:
There's a dirty commit here, and how can you reproduce the problem? An AR/AP entry must have a maturity date by definition. Also Travis is red.

manuelcalerosolis:
Step to reproduce:
Invoicing > Journal Entries
Create a new entrie and Post

image

pedrobaeza:
It's impossible to do that steps:

- Odoo doesn't allow you to post without lines.
- Odoo fills a maturity date automatically and doesn't allow you to empty it.

The only thing I can think about is a bad payment term without lines, but the less problem in that case is this error.

manuelcalerosolis:
It is not creating an invoice, it is creating an accounting entry.

pedrobaeza:
An accounting entry with a payable/receivable account? You shouldn't do that without maturity date as well. Anyway, it's OK to add that double check, but this PR still contains 2 commits, which is incorrect.

manuelcalerosolis:
This:

https://github.com/OCA/account-invoice-reporting/pull/170/commits/d66997312ad955449f581bc3ed85e76a139c64c6

pedrobaeza:
You have to fix the pull request having 2 commits.

account_invoice_report_due_list/models/account_move.py
```diff
@@ -31,10 +31,12 @@ def get_multi_due_list(self):
         if self.type in ["in_invoice", "out_refund"]:
             due_move_line_ids = self.line_ids.filtered(
                 lambda ml: ml.account_id.internal_type == "payable"
+                and ml.date_maturity is not False
```
pedrobaeza:
```suggestion
                and not ml.date_maturity
```

manuelcalerosolis: better:

and ml.date_maturity

done!

pedrobaeza (approved):

chusamo (approved):
Tested in local and LGTM

chusamo:
Is this how is suppose to see?
Captura de pantalla de 2021-03-14 10-12-01

PR #150, opened 2020-07-06
[13.0] [MIG] account_invoice_comment_template

feketemihai:

PR #174, opened 2021-02-16
[13.0][MIG] account_invoice_comment_template: Migration to 13.0

victoralmau:
Migration to 13.0

Please @joao-p-marques and @CarlosRoca13 can you review it?

Locked by:
- [x] https://github.com/OCA/reporting-engine/pull/504

@Tecnativa TT26758


feketemihai:
@victoralmau I think here you just have to check the diff from #150 , probably small migration script, but in the rest should work without the extra code in account.move.

victoralmau:
Thanks @feketemihai, i have changed it now about your initial ideas.
I have add you as co-author in commit migration.

joao-p-marques (approved):
Code and functional review :+1:

CarlosRoca13 (approved):

pedrobaeza:
Please include the migration scripts for recovering the comment fields from `account.invoice` to `account.move`. Example script to inspire on:

https://github.com/OCA/l10n-spain/blob/13.0/l10n_es_aeat_sii_oca/migrations/13.0.1.0.0/post-migration.py

victoralmau:
Migration scripts added.
In other similar addon `sale_comment_template` will be necessary too.

pedrobaeza:
@victoralmau please follow the example I have given to you, not other code, and avoiding ORM processing. Also take into account that this is for OpenUpgrade, not for regular updates (that are more debatable about the need of the library), so it's totally OK (and advised) to use `openupgradelib`.

account_invoice_comment_template/models/account_move.py
```diff
@@ -0,0 +1,71 @@
...
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    comment_template1_id = fields.Many2one(
```
feketemihai: Why do you still keep the code in here, isn't it easier to treat this in the migration script to include also partners. Like this you can remove the code, tests, views, translations...

victoralmau: What do you think about it @pedrobaeza ?

pedrobaeza: I think the approach is incorrect: we still need the specific custom texts on each document, as they can modify them.

victoralmau:
It's ready to merge right?
Ping @pedrobaeza

account_invoice_comment_template/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+# Copyright 2018 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Account Invoice Comments Template",
+    "summary": "Comments templates on invoice documents",
+    "version": "13.0.1.0.0",
+    "category": "Accounting & Finance",
+    "author": "Camptocamp, " "Tecnativa, " "Odoo Community Association (OCA)",
```
ernestotejeda:
```suggestion
    "author": "Camptocamp, Tecnativa, Odoo Community Association (OCA)",
```

account_invoice_comment_template/models/account_move.py
```diff
@@ -0,0 +1,71 @@
...
+        if comment:
+            self.note1 = comment.text
+
+    @api.onchange("comment_template2_id")
+    def _set_note2(self):
+        comment = self.comment_template2_id
+        if comment:
+            self.note2 = comment.text
+
+    @api.depends("partner_id")
```
ernestotejeda: Is it needed?

account_invoice_comment_template/models/account_move.py
```diff
@@ -0,0 +1,71 @@
...
+    def get_comment_template_item(self, position):
+        return self.env["base.comment.template"].search(
+            [
+                ("model_ids.model", "=", "account.move"),
+                ("position", "=", position),
+                ("partner_ids", "in", self.partner_id.ids),
+            ],
+            limit=1,
+        )
```
ernestotejeda: Shouldn't you take the invoice company into account and order by priority?, get_comment_template method of  comment.template model takes into account the company of the model and as i can see, base.comment.template model doesn't have the _order attribute defined.

account_invoice_comment_template/models/account_move.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @api.onchange("partner_id", "company_id")
+    def _onchange_partner_id(self):
+        res = super()._onchange_partner_id()
+        # before_lines
+        comment_template = self.get_comment_template_item("before_lines")
+        if comment_template:
+            self.comment_template1_id = comment_template
+            self._set_note1()
```
ernestotejeda: Is that function call necessary? I think the assignment of the previous line should already trigger the self_set_note1 method (since it is a onchange method tha depends on comment_template1_id), what do you think?

account_invoice_comment_template/models/account_move.py
```diff
@@ -0,0 +1,71 @@
...
+        # after_lines
+        comment_template = self.get_comment_template_item("after_lines")
+        if comment_template:
+            self.comment_template2_id = comment_template
+            self._set_note2()
```
ernestotejeda: The same here.

account_invoice_comment_template/tests/test_account_invoice_report.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_comments_in_invoice(self):
+        self.invoice._onchange_partner_id()
+        res = (
+            self.env["ir.actions.report"]
+            ._get_report_from_name("account.report_invoice")
+            .render_qweb_html(self.invoice.ids)
+        )
+        self.assertRegexpMatches(str(res[0]), self.before_comment.text)
+        self.assertRegexpMatches(str(res[0]), self.after_comment.text)
```
ernestotejeda: I think it's not responsibility of this module to test that a report prints its values correctly. Also, if this module is tested together with another module that hides those texts (via qweb) that are being tested here, then the test may fail.

account_invoice_comment_template/models/account_move.py
victoralmau: Thanks, i have changed to add `company_id` in domain and order by priority.
It's not possible to use `get_comment_template` function because need `base.comment.template` record and these function not return it. Maybe i have to create a little PR about it.

victoralmau: With these change https://github.com/OCA/reporting-engine/pull/497/ it's possible to change it something like these:

```
_name = "account.move"
_inherit = ["account.move", "comment.template"]

self.get_comment_template_records(
    position="before_lines", company_id=self.company_id.id, partner_id=self.partner_id.id
)[0]
```

ernestotejeda (approved):

account_invoice_comment_template/models/account_move.py
ernestotejeda: I think that would be good. I have already approved that PR.

account_invoice_comment_template/__manifest__.py
```diff
@@ -2,19 +2,21 @@
...
-    "category": "Sale",
-    "author": "Camptocamp, " "Tecnativa, " "Odoo Community Association (OCA)",
+    "version": "13.0.1.0.0",
+    "category": "Accounting & Finance",
+    "author": "Camptocamp, Tecnativa, Odoo Community Association (OCA)",
+    "development_status": "Mature",
     "license": "AGPL-3",
     "installable": True,
-    "depends": ["account", "base_comment_template",],
+    "depends": ["l10n_generic_coa", "base_comment_template"],
```
pedrobaeza: This dependency is not correct. We must do tests that don't force us to install a specific CoA.

account_invoice_comment_template/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,30 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO base_comment_template_ir_model_rel
+        (template_id, model_id)
+        SELECT bct.id, (
+            SELECT imd.id
```
pedrobaeza: Don't do this subquery, but join the table in the upper query.

account_invoice_comment_template/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,30 @@
...
+            FROM ir_model_data imd
+            WHERE imd.name = 'model_account_move'
+            AND imd.module = 'account'
+        ) AS model_id
+        FROM base_comment_template bct
+        WHERE bct.id IN (
+            SELECT comment_template1_id
```
pedrobaeza: Don't do subquery.

account_invoice_comment_template/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,30 @@
...
+            FROM account_move
+            WHERE comment_template1_id IS NOT NULL
+        ) OR bct.id IN (
+            SELECT comment_template2_id
```
pedrobaeza: Don't do subquery.

pedrobaeza (changes requested):

pedrobaeza:
This should be adapted now to the new many2many reality: the fields in the invoice should be also m2m, and filter them by company.

account_invoice_comment_template/models/account_move.py
```diff
@@ -0,0 +1,51 @@
...
+        ],
+    )
+    comment_template2_ids = fields.Many2many(
+        comodel_name="base.comment.template",
+        string="Bottom Comment Template",
+        domain=[
+            ("model_ids.model", "=", "account.move"),
+            ("position", "=", "after_lines"),
+        ],
+    )
```
joao-p-marques: I think you will need to give explicit names to this relations' columns, otherwise the ORM will just generate them from the model names, and as there are 2 relations between the same modules, it will cause the conflict in https://travis-ci.com/github/OCA/account-invoice-reporting/jobs/500445740#L1133

Also, as you will already need migration scripts, why not giving a more distinct name to these, like `comment_template_top_ids` and `comment_template_bottom_ids`?

joao-p-marques (changes requested):

pedrobaeza:
This PR must be changed to the latest changes on the base. Please check the base first and dismiss this one until the other is merged and this one adapted.

victoralmau:
According to 14.0, is correct rename addon to `account_comment_template`, right?

pedrobaeza:
Well, I don't see incorrect the current name, but if the renaming has been done, coming back is going to be awful.

victoralmau:
It's ready to review now.

account_invoice_comment_template/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        INSERT INTO account_move_base_comment_template_rel
+        (base_comment_template_id, account_move_id)
+        SELECT bct.id AS base_comment_template_id, am.id AS account_move_id
+        FROM account_move am
+        JOIN base_comment_template bct ON bct.id = am.old_comment_template2_ids
+        WHERE am.old_comment_template2_ids IS NOT NULL
+        """,
+    )
+    openupgrade.drop_columns(env, drop_columns)
```
pedrobaeza: Never drop columns. Remember: preservation principle. The only exception is if that data compromises integrity, which violates other principle. Even more, you will get an error, as they are m2m fields, not SQL columns. You can lift constraints in such tables for avoiding performance lost though.

account_invoice_comment_template/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+field_renames = [
```
pedrobaeza: This renaming is not preventing the m2m tables linking, so it doesn't serve. Even more, you don't want to rename fields, only columns (which is not the case), when renaming for preserving content in other place. The question here is: the m2m table has the same name? If that's the case, then you should rename the m2m table.

account_invoice_comment_template/models/account_move.py
```diff
@@ -0,0 +1,17 @@
...
+
+from odoo import fields, models
+
+
+class AccountMove(models.Model):
+    _name = "account.move"
+    _inherit = ["account.move", "comment.template"]
+
+    # Deprecated, these fields will be removed in the future (15.0 probably)
+    note1 = fields.Html(string="Top Comment", readonly=True)
```
pedrobaeza: Why not removing it? We can discuss in v14 for its removal, but I would remove it here for now.

pedrobaeza (changes requested):

account_invoice_comment_template/migrations/13.0.1.0.0/post-migration.py
victoralmau: Sorry for my mistake, the name of the fields was wrong (it is not an m2m, it is 2 m2o fields that convert to m2m).

pedrobaeza: You are still dropping columns and this is not the mentioned exception.

account_invoice_comment_template/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,35 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO account_move_base_comment_template_rel
+        (base_comment_template_id, account_move_id)
+        SELECT bct.id AS base_comment_template_id, am.id AS account_move_id
+        FROM account_move am
+        JOIN base_comment_template bct ON bct.id = am.old_comment_template2_id
+        WHERE am.old_comment_template2_id IS NOT NULL
+        """,
```
pedrobaeza, line 30: Add `ON CONFLICT DO NOTHING`, as there can be the case where you have the same template on top and on bottom.

account_invoice_comment_template/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+column_renames = {
+    "account_move": [
+        ("comment_template1_id", "old_comment_template1_id"),
```
pedrobaeza: Why are you renaming columns if the previous name is not overlapped by anything? The only reason for renaming columns is for avoiding to conflict with the same name, and as it's very unlikely that other module uses this field name, it's better to save on scripts doing the renaming + avoid to use the renamed one on further scripts. Anyways, if needed, the usual technique is to rename to `None`, for letting OpenUpgrade to put the name with `openupgradelib.get_legacy_name`.

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #175, opened 2021-02-18
[13.0] [MIG] account_invoice_report_payment_info: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT27461

please @carlosdauden @sergio-teruel review this


account_invoice_report_payment_info/models/account_invoice.py
```diff
@@ -0,0 +1,24 @@
...
+        info_pattern = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("account_invoice_report_payment_info.info_pattern", default="")
+        )
+        Payment = self.env["account.move.line"]
+        for payment_dict in res:
+            payment = Payment.browse(payment_dict["payment_id"])
+            payment_dict["move_ref"] = payment.move_id.ref
+            payment_dict["extra_info"] = info_pattern.format(**payment_dict)
```
victoralmau, line 23: What happen if extra_info is `False` 
IMO it's possible with a credit in customer invoice

```suggestion
            extra_info = info_pattern.format(**payment_dict)
            if extra_info: 
                payment_dict["extra_info"] = extra_info
```

CarlosRoca13: The extra_info deppends on the system parameter defined. With the option that you suggest and with the system parameter defined on data, the extra_info variable will take the value "(False)" so it will be evaluated as True. 

The problem of this, is that we don't know what will be defined on the system parameter, so it's not possible to avoid this problem.

account_invoice_report_payment_info/views/report_invoice.xml
```diff
@@ -5,7 +5,9 @@
         inherit_id="account.report_invoice_document_with_payments"
     >
         <xpath expr="//t[@t-foreach='payments_vals']/tr/td" position="inside">
-            <small t-esc="payment_vals['extra_info']" />
+            <t t-if="payment_vals['extra_info']">
```
victoralmau: It's really necessary to add `<t t-if `? (allways extra_info is defined according to `_get_reconciled_info_JSON_values` right?)

CarlosRoca13: Yes sorry, I set that to try to avoid what you said in last comment before investigate it, and i forget to delete it. Removed

victoralmau (approved):

account_invoice_report_payment_info/README.rst
```diff
@@ -0,0 +1,116 @@
...
+  * 'position': currency_id.position,
+  * 'date': payment.date,
+  * 'payment_id': payment.id,
+  * 'account_payment_id': payment.payment_id.id,
+  * 'invoice_id': payment.invoice_id.id,
+  * 'invoice_view_id': invoice_view_id,
+  * 'move_id': payment.move_id.id,
+  * 'ref': payment_ref,
+
+  https://github.com/odoo/odoo/blob/ceb0c985afbbb41951c6f953648a6d0795d4e537/addons/account/models/account_invoice.py#L185
```
sergio-teruel: This file not exists in v13.0

sergio-teruel (approved):
Little comment

sergio-teruel:
ping  @carlosdauden

account_invoice_report_payment_info/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,23 @@
...
+  * 'position': currency_id.position,
+  * 'date': payment.date,
+  * 'payment_id': payment.id,
+  * 'account_payment_id': payment.payment_id.id,
+  * 'invoice_id': payment.invoice_id.id,
+  * 'invoice_view_id': invoice_view_id,
+  * 'move_id': payment.move_id.id,
+  * 'ref': payment_ref,
+
+  https://github.com/odoo/odoo/blob/ceb0c985afbbb41951c6f953648a6d0795d4e537/addons/account/models/account_invoice.py#L185
```
carlosdauden: https://github.com/odoo/odoo/blob/1e35b8987c619f200e84da2ba97040b38347edde/addons/account/models/account_move.py#L1351
```suggestion
  https://github.com/odoo/odoo/blob/1e35b8987c619f200e84da2ba97040b38347edde/addons/account/models/account_move.py#L1351
```

carlosdauden (changes requested):

CarlosRoca13:
@sergio-teruel @carlosdauden change done

carlosdauden (approved):

PR #186, opened 2021-04-12
[14.0][MIG] account_invoice_report_hide_line

AmmarOfficewalaSerpentCS:

PR #189, opened 2021-04-21
[FIX] Fixed issued of a module not complitable with account_invoice_r…

AmmarOfficewalaSerpentCS:
…eport_hide_line.


AmmarOfficewalaSerpentCS:
#186

AmmarOfficewalaSerpentCS:
@bodedra , 

This is solution for a completable two module and 
Can suggest me better solution for this issue ?

pedrobaeza:
This is not a valid solution. A module should never query if other module is installed. You probably need to create a glue module for joining both.

PR #192, opened 2021-04-30
[FIX] account_invoice_report_grouped_by_picking: filter sections and notes

chienandalu:
Avoid nasty exceptions. In the other hand it would be hard to fit the
concept of sections and notes to the picking groups in which this report
results.

cc @Tecnativa TT29507

ping @sergio-teruel @pedrobaeza

### DIFF
--- a/account_invoice_report_grouped_by_picking/readme/ROADMAP.rst
+++ b/account_invoice_report_grouped_by_picking/readme/ROADMAP.rst
@@ -1,0 +1,4 @@
+* As a result of the grouping by pickings, we don't support notes and descriptions.
+* Anyways, an invoice with no pickings should be printed as usual. The problem is
+  that in the current state of the report, a heavy refactoring is needed to be able to
+  fallback to the regular behavior in such case.
--- a/account_invoice_report_grouped_by_picking/README.rst
+++ b/account_invoice_report_grouped_by_picking/README.rst
@@ -49,6 +49,14 @@ Usage
 #. Invoice report will group invoice lines and shows information about sales
    and pickings in every group.
 
+Known issues / Roadmap
+======================
+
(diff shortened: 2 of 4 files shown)


pedrobaeza (approved):
Please put a note in the ROADMAP about section and lines not being printed with this module.

chienandalu:
Yes. It's already there

account_invoice_report_grouped_by_picking/readme/ROADMAP.rst
```diff
@@ -0,0 +1,4 @@
+* As a result of the grouping by pickings we don't support notes and descriptions.
+  Anyway, an invoice with no pickings it should be printed as usual. The problem is
+  that in the current state of the report, we should refactor it to be able to
+  fallback to the regular behavior.
```
pedrobaeza:
```suggestion
* As a result of the grouping by pickings, we don't support notes and descriptions.
* Anyways, an invoice with no pickings should be printed as usual. The problem is
  that in the current state of the report, a heavy refactoring is needed to be able to
  fallback to the regular behavior in such case.
```

bodedra:
@pedrobaeza JFI, when I get time, I will work on glue module to support account_invoice_report_grouped_by_picking and account_invoice_report_hide_line module functionality. And @chienandalu will you do forward port for this fix?

account_invoice_report_grouped_by_picking/readme/ROADMAP.rst
chienandalu: Changed

chienandalu:
@bodedra yes, as soon as it's merged I'll forward it to 13.0

bodedra (approved):
Code review. :+1:

PR #165, opened 2021-01-08
[14.0][MIG] account_invoice_comment_template renamed to account_comment_template 

petrus-v:
@OCA/accounting-maintainers (I'm not sure it's a good practice to ping PSC like this ?)

As long `account.invoice` model were removed I'm wondering if it's good time to rename module from `account_invoice_comment_template` to `account_move_comment_template` ? Any thought welcomes!

~~This MR is based on #163~~
This MR depends on [base_commet_template migration](https://github.com/OCA/reporting-engine/pull/474)

PR #194, opened 2021-05-05
[14.0][IMP] account_comment_template: Include changes from 13.0

victoralmau:
Include changes from 13.0 to standardize both versions:

- Migration to v13: https://github.com/OCA/account-invoice-reporting/pull/174 + https://github.com/OCA/sale-reporting/pull/105#discussion_r632752570

Locked by:
- [x] https://github.com/OCA/reporting-engine/pull/500
- [x] https://github.com/OCA/reporting-engine/pull/509

Lock:
- [x] `sale_comment_template` https://github.com/OCA/sale-reporting/pull/113

Please @joao-p-marques and @pedrobaeza can you review it?

@Tecnativa TT29138

### DIFF
--- a/account_comment_template/models/account_move.py
+++ b/account_comment_template/models/account_move.py
@@ -2,51 +2,12 @@
 # Copyright 2013-2014 Nicolas Bessi (Camptocamp SA)
 # Copyright 2018 Vicent Cubells - Tecnativa
 # Copyright 2019 Iván Todorovich (Druidoo)
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import api, fields, models
+from odoo import models
 
 
-class AccountInvoice(models.Model):
-    _inherit = "account.move"
-
-    comment_template1_id = fields.Many2one(
-        "base.comment.template",
-        string="Top Comment Template",
-        domain=[("position", "=", "before_lines")],
-    )
-
-    comment_template2_id = fields.Many2one(
-        "base.comment.template",
-        string="Bottom Comment Template",
-        domain=[("position", "=", "after_lines")],
-    )
-
-    note1 = fields.Html("Top Comment")
-    note2 = fields.Html("Bottom Comment")
-
-    @api.onchange("comment_template1_id")
-    def _set_note1(self):
-        comment = self.comment_template1_id
-        if comment:
-            self.note1 = comment.get_value(self.partner_id.id)
-
-    @api.onchange("comment_template2_id")
-    def _set_note2(self):
-        comment = self.comment_template2_id
-        if comment:
-            self.note2 = comment.get_value(self.partner_id.id)
-
-    @api.onchange("partner_id", "company_id")
-    def _onchange_partner_id_set_templates_and_notes(self):
-        comment_template = self.partner_id.property_comment_template_id
-        if comment_template.position == "before_lines":
-            self.comment_template1_id = comment_template
-        elif comment_template.position == "after_lines":
-            self.comment_template2_id = comment_template
-        # Force note read when changing partner/company
-        # to update to the current partner language.
-        if self.comment_template1_id:
-            self._set_note1()
-        if self.comment_template2_id:
-            self._set_note2()
+class AccountMove(models.Model):
+    _name = "account.move"
+    _inherit = ["account.move", "comment.template"]
--- a/account_comment_template/tests/test_account_move_report.py
+++ b/account_comment_template/tests/test_account_move_report.py
@@ -1,44 +1,67 @@
 # Copyright 2017 Simone Rubino - Agile Business Group
 # Copyright 2018 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import Form, TransactionCase
 
 
 class TestAccountInvoiceReport(TransactionCase):
-    at_install = False
-    post_install = True
-
     def setUp(self):
-        super(TestAccountInvoiceReport, self).setUp()
+        super().setUp()
+        self.company = self.env.ref("base.main_company")
         self.base_comment_model = self.env["base.comment.template"]
+        self.move_obj = self.env.ref("account.model_account_move")
         self.before_comment = self._create_comment("before_lines")
         self.after_comment = self._create_comment("after_lines")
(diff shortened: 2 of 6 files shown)


account_comment_template/models/account_move.py
```diff
@@ -2,51 +2,11 @@
...
-            self.comment_template1_id = comment_template
-        elif comment_template.position == "after_lines":
-            self.comment_template2_id = comment_template
-        # Force note read when changing partner/company
-        # to update to the current partner language.
-        if self.comment_template1_id:
-            self._set_note1()
-        if self.comment_template2_id:
-            self._set_note2()
+    _inherit = ["account.move", "comment.template"]
```
joao-p-marques, line 13: This causes some issues when installing because of the double inheritance and the fact that you don't specify a name:
```suggestion
    _name = "account.move"
    _inherit = ["account.move", "comment.template"]
```
Also, you can rename the class to `AccountMove` to make it clearer.

joao-p-marques (commented):
Code and functional review

Apart from the comment, LGTM :+1:

joao-p-marques (approved):

victoralmau:
Not related error about `account_invoice_report_grouped_by_picking` addon.
I am going to investigate the error to try to resolve it.

account_comment_template/tests/test_account_move_report.py
```diff
@@ -1,44 +1,64 @@
...
+                "company_id": self.company.id,
                 "position": position,
                 "text": "Text " + position,
+                "model_ids": [(6, 0, self.move_obj.ids)],
             }
         )
 
-    def test_comments_in_invoice(self):
+    def test_comments_in_invoice_report(self):
+        self.invoice._compute_comment_template_ids()
```
joao-p-marques: This would not be necessary after the fix from https://github.com/OCA/reporting-engine/pull/509

It actually tricks the test and is missing a case that happens in the UI

account_comment_template/tests/test_account_move_report.py
```diff
@@ -1,44 +1,64 @@
...
         self.after_comment = self._create_comment("after_lines")
         self.partner = self.env["res.partner"].create({"name": "Partner Test"})
-        self.invoice_model = self.env["account.move"]
-        self.invoice = self.invoice_model.create(
+        self.partner.base_comment_template_ids = [
+            (4, self.before_comment.id),
+            (4, self.after_comment.id),
+        ]
+        self.income_account = self.env["account.account"].search(
+            [("user_type_id.name", "=", "Income")], limit=1
```
joao-p-marques: You should also filter by company here. The test user might have more than one company associated and this search will return an account belonging to a different company:
```suggestion
            [("user_type_id.name", "=", "Income"), ("company_id", "=", self.company.id)], limit=1
```

victoralmau:
Thanks @joao-p-marques , changes done.

pedrobaeza:
Travis is red

victoralmau:
Travis is green now.

joao-p-marques:
> Travis is green now. 

Nice :+1: 
However, as the problems seems to be that "replace by nothing" that is done in `account_invoice_report_grouped_by_picking`, I'm wondering if the best solution wouldn't be to remove that blind replace in that module and, instead, mark the field as invisible. It would certainly avoid having to set a custom priority in this module, and possibly in all other modules that hit this problem...

WDYT @victoralmau @pedrobaeza ?

pedrobaeza:
Yes, `replace` operations must be "replaced" by invisible or similar whenever possible.

victoralmau:
I have apply some little changes:
- Remove priority + change replace to invisible in report_invoice. `account_invoice_report_grouped_by_picking`
- Remove priority added previously to report_invoice in (is not needed now according to previous change) `account_comment_template`.

account_comment_template/views/report_invoice.xml
```diff
@@ -5,17 +5,24 @@
         id="report_invoice_document_comments"
         inherit_id="account.report_invoice_document"
     >
-
```
tafaRU, line 8: @victoralmau just noticed the`note1` and `note2` fields were restored in `14.0`(ref. [link](https://github.com/OCA/account-invoice-reporting/pull/165/commits/4d53d68f8a6f04657878a06255a766617355b2f4#diff-2f01dbf1fee3a3f5fd1717f36eb46d4e4617de5c0130e0b14519b52457402facR25-R26)).
Considering that they were removed in `13.0` as https://github.com/OCA/account-invoice-reporting/pull/174#discussion_r625996270 I think we should go back. Could you take the occasion to remove again them in this PR? What do you kink?

pedrobaeza: That fields are removed in this PR. Do you want them to be? There are several reasons for removing them in v13:

- Now you can add more than one template to be put in the invoices, added according sequence.
- They provoked a very large duplication of data with barely added value.
- People got confused if they modify the notes and then select again another template, losing that changes.

v14 has this because the migration was done directly from v12 to v14 without waiting for the v13 refactoring, but I think we should go the new way even in this version.

tafaRU: > That fields are removed in this PR.

Ok I see, I have misread the changes, sorry for the noise.

> I think we should go the new way even in this version.

Ok, let's go ahead.

tafaRU (approved):
Overall LGTM, thanks!
Just out of curiosity: why did you add https://github.com/OCA/account-invoice-reporting/pull/194/commits/486d1899e8fcb90cff370c8793437d9167f84eb8 in this PR?

victoralmau:
> Overall LGTM, thanks!
> Just out of curiosity: why did you add [486d189](https://github.com/OCA/account-invoice-reporting/commit/486d1899e8fcb90cff370c8793437d9167f84eb8) in this PR?

`account_invoice_report_grouped_by_picking` change is related to https://github.com/OCA/account-invoice-reporting/pull/194#issuecomment-848497157

tafaRU:
> `account_invoice_report_grouped_by_picking` change is related to [#194 (comment)](https://github.com/OCA/account-invoice-reporting/pull/194#issuecomment-848497157)

I meant the problem he is referring to in that comment. Anyway, no problem, I got it: https://travis-ci.com/github/OCA/account-invoice-reporting/jobs/508144419#L1370-L1377

Thanks a lot!

victoralmau:
> > `account_invoice_report_grouped_by_picking` change is related to [#194 (comment)](https://github.com/OCA/account-invoice-reporting/pull/194#issuecomment-848497157)
> 
> I meant the problem he is referring to in that comment. Anyway, no problem, I got it: https://travis-ci.com/github/OCA/account-invoice-reporting/jobs/508144419#L1370-L1377
> 
> Thanks a lot!

Thank you very much, you're right, now the changes to `account_invoice_report_grouped_by_picking` are not necessary (I have already removed them).

PR #197, opened 2021-05-25
[13.0][FIX] account_invoice_comment_template: Remove unnecessary line (in 14.0 it's an error)

victoralmau:
Remove unnecessary line (in 14.0 it's an error). Included in https://github.com/OCA/account-invoice-reporting/pull/194

Please @joao-p-marques and @pedrobaeza can you review it?

Locked by: 
- [x] `base_comment_template` https://github.com/OCA/reporting-engine/pull/512

@Tecnativa TT29138

### DIFF
--- a/account_invoice_comment_template/tests/test_account_invoice_report.py
+++ b/account_invoice_comment_template/tests/test_account_invoice_report.py
@@ -58,7 +58,6 @@ class TestAccountInvoiceReport(TransactionCase):
         )
 
     def test_comments_in_invoice_report(self):
-        self.invoice._compute_comment_template_ids()
         res = (
             self.env["ir.actions.report"]
             ._get_report_from_name("account.report_invoice")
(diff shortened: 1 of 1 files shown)


account_invoice_comment_template/tests/test_account_invoice_report.py
```diff
@@ -68,7 +68,6 @@ def test_comments_in_invoice_report(self):
         self.assertRegexpMatches(str(res[0]), self.after_comment.text)
 
     def test_comments_in_invoice(self):
-        self.partner.property_comment_template_id = self.after_comment.id
```
joao-p-marques, line 71: Shouldn't you assign `base_comment_template_ids` instead?
```suggestion
    def test_comments_in_invoice(self):
        self.partner.base_comment_template_ids = [(4, self.after_comment.id)]
```

joao-p-marques (commented):
Please check the comment, you will also need to change/delete https://github.com/OCA/reporting-engine/blob/14.0/base_comment_template/views/res_partner.xml right?

account_invoice_comment_template/tests/test_account_invoice_report.py
victoralmau: Is not necessary because they have already been defined in the setup test.

joao-p-marques (approved):

pedrobaeza (approved):

PR #201, opened 2021-07-02
[IMP] base_comment_template : allow user to set both(after and before lines) comment templates

chandni299:
On Partner, the user can set both types of comment templates.
1. a new field `property_comment_after_template_id` is added, to set only "After Line" comment template
2. Later, the Existing field `property_comment_template_id`, will require domain and context, so the user can set only the "Before Line" comment template. (It is not included in this PR, to keep existing logic)
3. Later it will require to adapt change to depended modules like "account_invoice_comment_template", "sale_comment_template" etc

### DIFF
--- a/base_comment_template/models/res_partner.py
+++ b/base_comment_template/models/res_partner.py
@@ -11,9 +11,14 @@ class ResPartner(models.Model):
         string='Conditions template',
         company_dependent=True,
     )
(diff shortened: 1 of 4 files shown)


benbrich (commented):
Thumbs up after functional test on local machine!

PR #206, opened 2021-09-03
[13.0][FIX] account_invoice_report_grouped_by_picking: Use Form in tests to avoid crash test

sergio-teruel:
Crash test due to this Odoo commit
https://github.com/OCA/OCB/commit/8ca10a8f39c6e6b8647a695aa543f008f01ba034
The wizard lines have the field uom_id related to move.product_uom and readonly=False, if you call directly to wiz.onchage_picking_id a write in stock move is executed.

cc @Tecnativa
ping @carlosdauden @CarlosRoca13


pedrobaeza (approved):

pedrobaeza:
Fast-tracking for not having failed CIs:

/ocabot merge patch

PR #208, opened 2021-09-26
[OU-IMP] account_invoice_comment_template: Improving migration script

etobella:
Without this change it took 15 minutes to process 600.000 invoices. Now it should take less than one minute. Similar approach to purchase from OpenUpgrade:

https://github.com/OCA/OpenUpgrade/blob/13.0/addons/purchase/migrations/13.0.1.2/pre-migration.py#L36
https://github.com/OCA/OpenUpgrade/blob/13.0/addons/purchase/migrations/13.0.1.2/post-migration.py#L32


pedrobaeza (approved):

pedrobaeza:
Please fix pre commit

PR #209, opened 2021-09-30
[14.0][MIG] account_invoice_line_sale_line_position: Migration to 14.0

nguyenhk99:


simahawk (approved):

account_invoice_line_sale_line_position/models/account_invoice.py
```diff
@@ -0,0 +1,34 @@
...
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    has_order_position = fields.Boolean(compute="_compute_has_order_position")
+
+    @api.depends("invoice_line_ids.position_formatted")
+    def _compute_has_order_position(self):
+        for record in self:
+            record.has_order_position = any(
```
simahawk, line 15: it would be better to loop on each line and stop at the 1st one having the value. Not a blocker tho,

TDu (approved):

PR #212, opened 2022-01-12
[MIG] account_invoice_comment_template: Migration to 14.0

olgamarcocb:
Standard Migration


emagdalenaC2i:
This is already merged and moved to 
account-invoice-reporting in https://github.com/OCA/account-invoice-reporting/pull/165

See https://github.com/OCA/account-invoice-reporting/issues/154

PR #213, opened 2022-02-10
[14.0][IMP] account_invoice_report_grouped_by_picking: No display repeated lines for refund invoices without no returned quantities

ferran-S73:
Brings back https://github.com/OCA/account-invoice-reporting/pull/171 which was closed due to inactivity.

PR #214, opened 2022-02-10
[FIX] account_invoice_report_grouped_by_picking: unexpected lines when printing refund

oihane:


oihane:
Ups, there is another PR with same issue. I'll close this

#213

PR #216, opened 2022-02-17
[13.0][IMP] account_invoice_report_due_list: Add compute_sudo=True to avoid errors

victoralmau:
Add `compute_sudo=True` to avoid errors if the user does not have access to the account.

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT30686

### DIFF
--- a/account_invoice_report_due_list/models/account_move.py
+++ b/account_invoice_report_due_list/models/account_move.py
@@ -9,7 +9,9 @@ class AccountMove(models.Model):
     _inherit = "account.move"
 
     multi_due = fields.Boolean(string="Multiple date due", compute="_compute_multi_due")
-    multi_date_due = fields.Char(string="Due Dates", compute="_compute_multi_date_due")
+    multi_date_due = fields.Char(
+        string="Due Dates", compute="_compute_multi_date_due", compute_sudo=True
+    )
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Possible record rules over entries may cause a problem, so let's do this preventively as there's no side effect on it

carlosdauden (approved):

account_invoice_report_due_list/models/account_move.py
```diff
@@ -9,7 +9,9 @@ class AccountMove(models.Model):
     _inherit = "account.move"
 
     multi_due = fields.Boolean(string="Multiple date due", compute="_compute_multi_due")
```
pedrobaeza, line 11: Sorry, this is the one needing the compute_sudo.

pedrobaeza (changes requested):

account_invoice_report_due_list/models/account_move.py
victoralmau: Sorry, but I think that finally it is not necessary to make any changes in this addon.

pedrobaeza: Yeah, sure. I was thinking in the same side effect on restricting accounts, but I'm thinking the problem will be in other place then.

PR #220, opened 2022-05-06
[13.0][ADD] account_reporting_volume: New module to show the volume in the invoices analysis view

CarlosRoca13:
cc @Tecnativa TT36336

please @victoralmau @sergio-teruel


pedrobaeza:
Shouldn't it be called `account_invoice_pivot_volume` (`account_invoice_report_volume` may be misleading)?

CarlosRoca13:
IDK, I take as reference this other: https://github.com/OCA/account-invoice-reporting/tree/13.0/account_reporting_weight

pedrobaeza:
I see. There's no good name, as core uses `report` for both pivot and QWeb. I suppose it was put `reporting` for differentiating this module from the existing one modifying the report. In other places, we have use the word `pivot` for such differentiation, but then it wouldn't be homogeneous with the existing, so OK to keep it.

sergio-teruel (approved):

pedrobaeza (approved):

PR #222, opened 2022-05-24
[13] add partner id name when parent id

sergioeix:
When you do a group invoice and the partner has a parent now this info is shown. Is usefull as example if you have a customer with 5 shipping addres and you invoice him to 1 invoice contact.

### DIFF
--- a/account_invoice_report_grouped_by_picking/views/report_invoice.xml
+++ b/account_invoice_report_grouped_by_picking/views/report_invoice.xml
@@ -46,6 +46,12 @@
                                 t-options="{'widget': 'date'}"
                             />
                             <span t-field="picking.name" />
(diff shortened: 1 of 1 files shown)


sergioeix:
Hi this is a small change, can you review please.
@pedrobaeza 
@chienandalu 
@sergio-teruel 
@carlosdauden

sergio-teruel:
@sergioeix I think that this change is not needed... I think that this is much info for picking line..

sergioeix:
It only adds the info when needed only when a parent is set, this info is
for the customer facility to check the invoice.

pedrobaeza:
It can be OK if you check if the partner is different from the invoice, and only in that case, you put such info.

PR #227, opened 2022-07-03
[12.0][FIX] account_reporting_weight: weight is always zero

cubells:
Fix query to create view in a correct manner.

### DIFF
--- a/account_reporting_weight/readme/INSTALL.rst
+++ b/account_reporting_weight/readme/INSTALL.rst
@@ -1,2 +1,0 @@
-* `product_weight_through_uom` is recommended in case you have products with
-  a UoM in the Weight category so the weight is correctly computed.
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why the factors would be `NULL`? There's always both UoMs. And also don't remove the RST file.

cubells:
SQL must be altered by some changes after the module was published and now module does not work. See:

image


With these changes, all works now again.

I restore file again, but one question:

Do you want to not delete the file by historical reasons or for what? For me content of this file make no sense now.

pedrobaeza:
> SQL must be altered by some changes after the module was published and now module does not work. See:

Isn't that due to bad demo data? Is this happening in real DBs? `factor` and `factor2` shouldn't be `NULL` having both uom (the product and the invoice ones). That's why I'm wondering why it happens.

> Do you want to not delete the file by historical reasons or for what? For me content of this file make no sense now.

Isn't that recommendation still valid?

cubells:
You are right @pedrobaeza , in a production database, weight is calculated.

So I close this PR.

Thanks

PR #239, opened 2022-10-03
[MIG/IMP] Migrate partner time to pay module from v13.0 to v14.0

Urvisha-OSI:


rafaelbn:
Already migrated https://github.com/OCA/account-invoice-reporting/tree/14.0/partner_time_to_pay

PR #241, opened 2022-10-17
[13.0][FIX] account_invoice_report_grouped_by_picking: Duplicated invoice lines

CarlosRoca13:
After the patch on stock_picking_invoice_link to link the refund moves to the return pickings created before the refund. This module shows on the picking duplicated lines.

With this patch the sign is correctly set and this duplication does not exist.

Before this patch:
imagen

After this patch:
imagen

cc @Tecnativa TT39916

please @pedrobaeza @sergio-teruel review this


sergio-teruel (approved):

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to upper versions

PR #243, opened 2022-10-27
[14.0][FIX] account_invoice_report_grouped_by_picking: Duplicated invoice lines

CarlosRoca13:
After the patch on stock_picking_invoice_link to link the refund moves to the return pickings created before the refund. This module shows on the picking duplicated lines.

With this patch the sign is correctly set and this duplication does not exist.

FW-PORT of https://github.com/OCA/account-invoice-reporting/pull/241

@Tecnativa TT39916 

please @sergio-teruel @victoralmau review this


ioans73 (approved):
LGTM 👍🏻

victoralmau (approved):

PR #245, opened 2022-11-08
[13.0][FIX] account_invoice_report_grouped_by_picking: refund invoice report

chienandalu:
Consider pending pickings that are linked to the invoice.

cc @Tecnativa TT40412

please review @sergio-teruel @carlosdauden

### DIFF
--- a/account_invoice_report_grouped_by_picking/models/account_move.py
+++ b/account_invoice_report_grouped_by_picking/models/account_move.py
@@ -15,7 +15,10 @@ class AccountMove(models.Model):
     @api.model
     def _sort_grouped_lines(self, lines_dic):
         return sorted(
-            lines_dic, key=lambda x: (x["picking"].date, x["picking"].date_done)
+            lines_dic,
+            key=lambda x: (
+                (x["picking"].date, x["picking"].date_done or x["picking"].date)
+            ),
         )
 
     def _get_signed_quantity_done(self, invoice_line, move, sign):


account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -15,7 +15,12 @@ class AccountMove(models.Model):
     @api.model
     def _sort_grouped_lines(self, lines_dic):
         return sorted(
-            lines_dic, key=lambda x: (x["picking"].date, x["picking"].date_done)
+            lines_dic,
+            key=lambda x: (
+                x["picking"].date_done
+                and (x["picking"].date, x["picking"].date_done)
+                or (x["picking"].date, x["picking"].date_done)
```
pedrobaeza: Isn't this the same as the previous line?

chienandalu: Sure, it wasn't the intention. Thanks :S

chienandalu: Fixed!

pedrobaeza (approved):

account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -15,7 +15,12 @@ class AccountMove(models.Model):
     @api.model
     def _sort_grouped_lines(self, lines_dic):
         return sorted(
-            lines_dic, key=lambda x: (x["picking"].date, x["picking"].date_done)
+            lines_dic,
+            key=lambda x: (
+                x["picking"].date_done
+                and (x["picking"].date, x["picking"].date_done)
+                or (x["picking"].date, x["picking"].date)
+            ),
```
carlosdauden:
```suggestion
            lines_dic, key=lambda x: (x["picking"].date, x["picking"].date_done or x["picking"].date)
```

account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -15,7 +15,12 @@ class AccountMove(models.Model):
     @api.model
     def _sort_grouped_lines(self, lines_dic):
         return sorted(
-            lines_dic, key=lambda x: (x["picking"].date, x["picking"].date_done)
+            lines_dic,
+            key=lambda x: (
+                x["picking"].date_done
+                and (x["picking"].date, x["picking"].date_done)
+                or (x["picking"].date, x["picking"].date)
```
carlosdauden: Another option:
```suggestion
                or x["picking"].date
```

carlosdauden (changes requested):

account_invoice_report_grouped_by_picking/models/account_move.py
chienandalu: This one wouldn't be valid as we would have a mix of tuples and datetimes.

account_invoice_report_grouped_by_picking/models/account_move.py
chienandalu: I went with this suggestion

carlosdauden (approved):

PR #254, opened 2022-11-30
[15.0][FIX] account_comment_template: Set models field in templates created from tests.

victoralmau:
Set models field in templates created from tests (since https://github.com/OCA/reporting-engine/pull/677).

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa


pedrobaeza (approved):

pedrobaeza:
We can fast-track this one as straight-forward and being for tests:

/ocabot merge patch

PR #255, opened 2022-12-02
[15.0][MIG] account_invoice_report_grouped_by_picking: MIgratio to 15.0

CarlosRoca13:
cc @Tecnativa TT36422

This PR depends of:

- [x] https://github.com/OCA/stock-logistics-workflow/pull/1147

This PR supersed https://github.com/OCA/account-invoice-reporting/pull/235

ping @carlosdauden @sergio-teruel

PR #256, opened 2023-01-03
[MIG] account_invoice_report_grouped_by_picking: 16.0

ferran-S73:
Migration from https://github.com/OCA/account-invoice-reporting/pull/255


account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -0,0 +1,100 @@
...
+    _inherit = "account.move"
+
+    @api.model
+    def _sort_grouped_lines(self, lines_dic):
+        return sorted(
+            lines_dic,
+            key=lambda x: (
+                (x["picking"].date, x["picking"].date_done or x["picking"].date)
+            ),
+        )
```
hailangvn: When mixing with invoice line that doesn't have neither delivery nor sale order line, this code caused error.

```
  File "/opt/OCA/account-invoice-reporting/tree/16.0/account_invoice_report_grouped_by_picking/models/account_move.py", line 104, in lines_grouped_by_picking
    return no_picking + self._sort_grouped_lines(with_picking)
  File "/opt/OCA/account-invoice-reporting/tree/16.0/account_invoice_report_grouped_by_picking/models/account_move.py", line 18, in _sort_grouped_lines
    return sorted(
TypeError: '<' not supported between instances of 'bool' and 'datetime.datetime'

The above exception was the direct cause of the following exception:

odoo.addons.base.models.ir_qweb.QWebException: Error while render the template
TypeError: '<' not supported between instances of 'bool' and 'datetime.datetime'
Template: account.report_invoice_document
Path: /t/t/div[2]/div[1]/table/tbody/t[3]
Node: <t t-foreach="o.lines_grouped_by_picking()" t-as="lines_group"/>
```

IMHO, there should be better solution than this.
```suggestion
        DTF = '%Y-%m-%d %H:%M:%S'
        return sorted(
            lines_dic,
            key=lambda x: (
                x["picking"] and (
                    x["picking"].date and x["picking"].date.strftime(DTF),
                    (x["picking"].date_done or x["picking"].date).strftime(DTF),
                ) or ('', '')
            ),
        )
```

account_invoice_report_grouped_by_picking/tests/test_account_invoice_group_picking.py
```diff
@@ -0,0 +1,294 @@
...
+        self.sale2.picking_ids[:1].move_line_ids.write({"qty_done": 1})
+        self.sale2.picking_ids[:1]._action_done()
+        sales = self.sale | self.sale2
+        # invoice sales
+        invoice = sales._create_invoices()
+        # Test directly grouping method
+        groups = invoice.lines_grouped_by_picking()
+        self.assertEqual(len(groups), 4)
+        self.assertEqual(groups[0]["picking"], groups[1]["picking"])
+        self.assertEqual(groups[2]["picking"], groups[3]["picking"])
```
hailangvn, line 94: This is the test to generate the error above.

```suggestion
        self.assertEqual(groups[2]["picking"], groups[3]["picking"])
        # mix with invoice line that doesn't have neither move_line_ids
        # nor sale_line_ids
        vals = [
            {
                "name": "test invoice line",
                "move_id": invoice.id,
                "price_unit": 10.0,
                "account_id": invoice.invoice_line_ids[0].account_id.id,
            }
        ]
        lines = invoice.invoice_line_ids.create(vals)
```

hailangvn (commented):
Thanks for the work. I have done functional test and the PR is good except for one scenario. Please consider it.

ferran-S73:
Hi @hailangvn I agree with you in that this fix is needed and I'm going to add your changes but I don't think it's a good idea to force the datetime format since other languages may have a different one, I think it's better to let Odoo handle it via the language of the report

ferran-S73:
Hi again @hailangvn I added your suggestions in a new commit to the module. For the date format part I try to construct the format based on the partner's language in order to sort by strings. Thanks for the suggestions!

hailangvn:
Thanks @ferran-S73 for applying the fix. I think the purpose of date format is to sort so the ISO one is good enough. The format from partner's language has the purpose of displaying and can cause different behavior when sorting. For example:

```
>>> from datetime import datetime
>>> a = [datetime(2023, 1, 17, 14, 59), (2023, 1, 18, 9, 10)]
>>> lang = self.env["res.lang"].search([("code", "=", 'en_US')])
>>> DTF = "{} {}".format(lang.time_format, lang.date_format)
>>> DTF
'%H:%M:%S %d.%m.%Y'
>>> sorted(a, key=lambda x: x.strftime(DTF))
[datetime.datetime(2023, 1, 18, 9, 10), datetime.datetime(2023, 1, 17, 14, 59)]
>>> ISO = "%Y-%m-%d %H:%M:%S"
>>> sorted(a, key=lambda x: x.strftime(ISO))
[datetime.datetime(2023, 1, 17, 14, 59), datetime.datetime(2023, 1, 18, 9, 10)]
```

account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -53,7 +53,9 @@ def lines_grouped_by_picking(self):
         # Let's get first a correspondance between pickings and sales order
         so_dict = {x.sale_id: x for x in self.picking_ids if x.sale_id}
         # Now group by picking by direct link or via same SO as picking's one
-        for line in self.invoice_line_ids.filtered(lambda x: not x.display_type):
+        for line in self.invoice_line_ids:
+            if line.display_type != "product":
```
ivantodorovich, line 68: Shouldn't we also process consumables here ?

```py
if line.display_type not in ("product", "consu"):
```

ivantodorovich (approved):

ferran-S73:
> Thanks @ferran-S73 for applying the fix. I think the purpose of date format is to sort so the ISO one is good enough. The format from partner's language has the purpose of displaying and can cause different behavior when sorting. For example:
> 
> ```
> >>> from datetime import datetime
> >>> a = [datetime(2023, 1, 17, 14, 59), (2023, 1, 18, 9, 10)]
> >>> lang = self.env["res.lang"].search([("code", "=", 'en_US')])
> >>> DTF = "{} {}".format(lang.time_format, lang.date_format)
> >>> DTF
> '%H:%M:%S %d.%m.%Y'
> >>> sorted(a, key=lambda x: x.strftime(DTF))
> [datetime.datetime(2023, 1, 18, 9, 10), datetime.datetime(2023, 1, 17, 14, 59)]
> >>> ISO = "%Y-%m-%d %H:%M:%S"
> >>> sorted(a, key=lambda x: x.strftime(ISO))
> [datetime.datetime(2023, 1, 17, 14, 59), datetime.datetime(2023, 1, 18, 9, 10)]
> ```

@hailangvn you're right, sorry I got confused thinking these dates would get printed in the report. I applied your fix

hailangvn (approved):

gurneyalex (approved):

damdam-s:
issue w/ isort version in pre-commit config

```                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            File "/tmp/pip-build-env-xhz3j2o6/overlay/lib/python3.11/site-packages/poetry/core/factory.py", line 57, in create_poetry
              raise RuntimeError("The Poetry configuration is invalid:\n" + message)
          RuntimeError: The Poetry configuration is invalid:
            - [extras.pipfile_deprecated_finder.2] 'pip-shims<=0.3.4' does not match '^[a-zA-Z-_.0-9]+$'
            - ```

PR #226, opened 2022-07-03
[ADD][12.0] account_reporting_net_weight: new module

cubells:

PR #263, opened 2023-01-23
Add copier

SirTakobi:
See https://github.com/OCA/oca-addons-repo-template


SirTakobi:
@OCA/accounting-maintainers can someone have a look? This is blocking all the `12.0` CIs
@pedrobaeza I just noticed this is what you asked for in https://github.com/OCA/account-invoice-reporting/pull/226#issuecomment-1304800597

PR #268, opened 2023-03-20
[15.0][FIX] partner_time_to_pay: Data is calculated with partner + partner.child_ids invoices

Shide:
The data is calculated using the contact and its children, so in each contact contact the aggregated data can be consulted.

Before there was no way to get what the statistics were at the Company level because the number of invoices was not in any field and a true proportion could not be made.
The data exclusively from the father and the children did not make sense because normally the information that is going to be used later (to provide information to investors, calculate a better payment term, etc.) requires knowing the data at the Company level in general (and perhaps for each Contact separately).

Now you have a vision of how the Company pays in general and how each of its children pays specifically.

MT-2393 @moduon @rafaelbn @max3903

### DIFF
--- a/partner_time_to_pay/models/res_partner.py
+++ b/partner_time_to_pay/models/res_partner.py
@@ -39,10 +39,10 @@ class ResPartner(models.Model):
     def _compute_d2x(self):
         for partner in self:
             partner.d2r_ytd, partner.d2r_life = partner._compute_d2x_per_invoice_type(
-                partner.invoice_ids, {"out_invoice"}
+                (partner + partner.child_ids).invoice_ids, {"out_invoice"}
             )
             partner.d2p_ytd, partner.d2p_life = partner._compute_d2x_per_invoice_type(
-                partner.invoice_ids, {"in_invoice"}
+                (partner + partner.child_ids).invoice_ids, {"in_invoice"}
             )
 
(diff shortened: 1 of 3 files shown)


loida-vm (approved):
Functional review done!
LGTM 👌

Only one think, in the list view of the invoices I can't see the date. You can see in the screenshot
image

Shide:
The field you are lookin for is the Payment date (The date when it's fully paid).
The module calculates the date between (move date and payment date). Payment date is calculated when a full reconcile is done in the invoice.

loida-vm:
> The field you are lookin for is the Payment date (The date when it's fully paid). The module calculates the date between (move date and payment date). Payment date is calculated when a full reconcile is done in the invoice.

Right! Everything is OK!

ArantxaSudon:
Functional Review Done 👌

rafaelbn (approved):
+1

@max3903 ? We have the problem that we didn't have information in commercial partner as usual -> https://twitter.com/alexisdlattre/status/1618377832090185730?t=RXod5yqdkWjAPcvL979QXQ&s=19

ArantxaSudon (approved):

fcvalgar (approved):
LGTM, thank you @Shide

PR #270, opened 2023-04-12
[16.0][MIG] account_invoice_report_payment_info: Migration to version 16.0

stefan-tecnativa:
cc @Tecnativa TT42418

@CarlosRoca13 please review!

ping @sergio-teruel


account_invoice_report_payment_info/models/account_invoice.py
```diff
@@ -0,0 +1,24 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.move"
+
+    def _get_reconciled_info_JSON_values(self):
+        res = super()._get_reconciled_info_JSON_values()
```
CarlosRoca13: This method does not exist anymore. So this is no having effect. 

Check where is the data collected on the report to replicate the behaviour.

CarlosRoca13 (changes requested):

CarlosRoca13 (commented):
It works, but I'm not sure if the code is achieving the same result as in the latest versions.

Can you confirm @carlosdauden

sergio-teruel (approved):
Functional and code review ok

sergio-teruel:
ping @juangomezzurita

CarlosRoca13 (approved):

PR #278, opened 2023-06-23
[13.0][FIX] account_invoice_report_grouped_by_picking: sold service refund

chienandalu:
If we're trying to make a refund of a service that had a link to a sale line those lines weren't ever returned.

cc @Tecnativa TT44067

please review @carlosdauden


carlosdauden (approved):

carlosdauden (changes requested):
xxx

chienandalu (commented):
Thanks @carlosdauden for your amend :smile: 
Thanks @stefan-tecnativa for your test case :+1:

pedrobaeza:
@stefan-tecnativa please fw-port it to upper versions:

/ocabot merge patch

PR #286, opened 2023-08-21
[16.0][IMP] account_comment_template: allow to use dynamic information in template

alan196:
With this change it allows to generate templates with dynamic content


pedrobaeza (changes requested):
- `t-raw` is deprecated. Use `t-out`. Anyway, that's not for dynamic expressions, but for HTML.
- This is not `[ADD]`, but `[IMP]`.

alan196:
@pedrobaeza thanks for the review! I updated the requested changes 👍

pedrobaeza (approved):

YahairaCVJarsa (approved):

PR #293, opened 2023-10-18
[16.0][FIX] account_invoice_report_payment_info: compute should be multi

yajo:
super() is multi-record, like all compute methods.

This one should be too. Otherwise, it fails to integrate with `l10n_fr_account_vat_return` with `ValueError: Expected singleton: account.move(739, 737, 741, 735)`.

@moduon MT-1075


pedrobaeza (approved):

rafaelbn (approved):
/ocabot merge patch

PR #300, opened 2023-12-22
[14.0][FIX] account_invoice_report_grouped_by_picking: test compatibility

chienandalu:
The module `sale_order_line_sequence` adds a column to the invoice report with a sequence that could contain the name of the sales order when the lines belong to different orders. Let's make sure that we count only our row content.

cc @Tecnativa TT46702

please review @CarlosRoca13 @carlosdauden


pedrobaeza:
Rebase after we have merged #298

pedrobaeza:
Are tests failing due to the order of the lines?

chienandalu:
I don't know really. This PR doesn't change that much. I'm retrying to see if its a random quirk

chienandalu:
Now one passed and the other failed :thinking:

PR #303, opened 2024-03-06
[16.0][FIX]account_invoice_report_grouped_by_picking: do not show negative values with partially invoiced services

manuelregidor:
With this fix, partially invoiced services that come from sale order lines do not show negative values in invoices PDFs grouped by pickings.

Consider the following sale order:
grouped_1

and the following invoice, in which the amount of the invoiced service quantity is less than the quantity ordered:
grouped_2

Before this fix, there were two lines for the service product: in one of them, quantity is 5, which is the amount in the sale order; in the other one, quantity is -3, which is the amount that needs to be substracted from the other line in order to obtain the amount present in the invoice:
grouped_3

With the fix, there is only one line in the PDF that contains the amount that is actually invoiced.
grouped_4


manuelregidor:
@HaraldPanten

HaraldPanten (approved):
Functional review. LGTM

etobella (approved):
Seems logic, can you add a test?

marcelsavegnago (approved):

CarlosRoca13 (approved):

pedrobaeza (changes requested):
Sorry for stopping this, but I see that your change impacts both lines with product, consumable or service, so this may alter some of the previous behaviors. Should you limit it to services?

Also I would like to have a test exercising the specific problem for avoiding future regressions.

account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -108,6 +108,10 @@ def lines_grouped_by_picking(self):
                         )
                         picking_dict.setdefault(key, 0)
                         qty = so_line.product_uom_qty
+                        if remaining_qty < so_line.product_uom_qty:
```
pedrobaeza: and product type = service?

manuelregidor: I think the same logic should be followed for all products.

Take this sale order as an example:
group_1

This picking:
group_2

And this invoice:
group_3

This is the PDF:
group_4

The issue happens as well, as the amount in the invoice is less than the amount in the sale order plus the invoice line is not related to any stock.move.line as it has not been delivered.

pedrobaeza: @carlosdauden I think you implemented this system with negatives. Do you remember why?

HaraldPanten: @manuelregidor is right, we should consider services and consumables as well, as the same is happening in both situations.

pedrobaeza: Well, this is considering everything (including storable ones). That's why I'm asking my colleague if he remembers about this implementation, as I think he was the one doing it.

HaraldPanten: Yes, I said consumables as the example was with a consumable product, but the same is happening with any type of product. See this example:

Sale order with storable products:

Captura de Pantalla 2024-03-08 a las 9 50 34

Partial delivery order:

Captura de Pantalla 2024-03-08 a las 9 50 46

Invoicing the delivered products and just some of the other storables (invoicing by ordered quantities):

Captura de Pantalla 2024-03-08 a las 9 50 57

The report is strange like that:

Captura de Pantalla 2024-03-08 a las 9 51 10

HaraldPanten:
Hi @pedrobaeza Any news about this? The error is in several versions, and it should be fixed.

pedrobaeza:
@carlosdauden can you confirm that you see the change correct?

account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -108,6 +108,10 @@ def lines_grouped_by_picking(self):
...
+                            qty = so_line.product_uom_qty - (
+                                so_line.product_uom_qty - remaining_qty
+                            )
```
pedrobaeza: Matemáticamente, esto es lo mismo:

```suggestion
                            qty = remaining_qty
```

manuelregidor: done

account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -108,6 +108,8 @@ def lines_grouped_by_picking(self):
                         )
                         picking_dict.setdefault(key, 0)
                         qty = so_line.product_uom_qty
+                        if remaining_qty < so_line.product_uom_qty:
+                            qty = remaining_qty
```
pedrobaeza: And what about this more compact code?

```suggestion
                        qty = min(so_line.product_uom_qty, remaining_qty)
```

manuelregidor: done

pedrobaeza (approved):
OK, let's go with this with the current consensus, and let's see if any problem is detected:

/ocabot merge patch

PR #258, opened 2023-01-12
[13.0][IMP] account_invoice_report_grouped_by_picking: Show section and notes

carlosdauden:
- Compatibility with section and notes lines (display_type)
 - Remove roadmap
 - Fix last_picking position to avoid picking header repeat in first line.

@Tecnativa TT39052

PR #308, opened 2024-04-09
[14.0][IMP] account_invoice_report_grouped_by_picking: Show section and notes

manuelregidor:
From https://github.com/OCA/account-invoice-reporting/pull/258

Commit 15363c54813bb186df4b9a5b572ff17f68b2dbaa applies changes in tests in module account_invoice_report_grouped_by_picking_sale_mrp because of method _sort_grouped_lines in account.move in module  account_invoice_report_grouped_by_picking. When this method is called in test_account_invoice_group_picking in module account_invoice_report_grouped_by_picking_sale_mrp, it is suposed to get a list of picking in a certain order. However, as the creation and validation of pickings in the test is done so quickly, the fields date and date_done in the pickings are the same, so the funtion _sort_grouped_lines does no sorting and the test fails.

T-5666

### DIFF
--- a/account_invoice_report_grouped_by_picking/README.rst
+++ b/account_invoice_report_grouped_by_picking/README.rst
@@ -52,14 +52,6 @@ Usage
 #. Invoice report will group invoice lines and show information about sales
    and pickings in every group.
 
-Known issues / Roadmap
-======================
-
-* As a result of the grouping by pickings, we don't support notes and descriptions.
-* Anyways, an invoice with no pickings should be printed as usual. The problem is
-  that in the current state of the report, a heavy refactoring is needed to be able to
-  fallback to the regular behavior in such case.
-
 Bug Tracker
 ===========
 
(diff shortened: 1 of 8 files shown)


manuelregidor:
@Jesarregui @ValentinVinagre @HaraldPanten

HaraldPanten:
@sergio-teruel @pedrobaeza That might be interesting for you. THX for your comments.

pedrobaeza:
Well, we don't have any customer in this version with this module. Things I see in a quick review:

- Attribution hasn't been done on the original commit.
- Why the change to put it at the end of the invoice?

Jesarregui (approved):
LGTM

manuelregidor:
@pedrobaeza 

I have added the only difference in attributions I have found in commit https://github.com/OCA/account-invoice-reporting/pull/258 (which is in file account_move.py). Please, let me know if I'm missing something.

The commit related to notes and sections at the end of the invoice has been applied as we noticed that those notes and sections were not shown with the origial commit. So, for instance, these are the lines in an invoice:

Section A
Note A
Product Line
Section B
Note B

Without the addition commit in this PR, Section B and Note B would not be shown in the PDF.

pedrobaeza:
OK, seems correct

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):

PR #318, opened 2024-05-07
[15.0][FIX] account_invoice_report_grouped_by_picking: Avoid crash on empty groups

pedrobaeza:
- Create an invoice with one line without product
- Print
- You get the following error:

AttributeError: 'bool' object has no attribute 'get'

Error when render the template
AttributeError: 'bool' object has no attribute 'get' Template: account.report_invoice_document
Path: /t/t/div/table/tbody/t[4]/tr[2]
Node: <tr t-if="lines_group.get('is_last_section_notes') and (not next_line or not next_line[0].get('is_last_section_notes')) and subtotal"/>

@Tecnativa TT49057

### DIFF
--- a/account_invoice_report_grouped_by_picking/views/report_invoice.xml
+++ b/account_invoice_report_grouped_by_picking/views/report_invoice.xml
@@ -126,7 +126,7 @@
         <!-- Append subtotal after notes and sections at the end of the invoice-->
         <xpath expr="//t[@name='account_invoice_line_accountable']/.." position="after">
             <tr
-                t-if="lines_group.get('is_last_section_notes') and (not next_line or not next_line[0].get('is_last_section_notes')) and subtotal"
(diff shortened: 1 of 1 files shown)


manuelregidor (approved):
Technical review. LGTM

carlosdauden (approved):

rafaelbn:
Hello @pedrobaeza , we did another fix https://github.com/OCA/account-invoice-reporting/pull/320 in 16 maybe is related. FYI 😄

pedrobaeza:
@manuelregidor I think you can fw-port to v17 the whole features taking into account the needed patches, and taking the approach of #320

HaraldPanten:
> @manuelregidor I think you can fw-port to v17 the whole features taking into account the needed patches, and taking the approach of #320

We'll do that. Don't worry.

PR #320, opened 2024-05-16
[FIX] [16.0] account_invoice_report_grouped_by_picking: next_line empty to avoid dict.get

Shide:
next_line is empty to avoid dict.get and throw an AttributeError

MT-6181 @moduon @EmilioPascual @rafaelbn @fcvalgar @Gelojr please review if you want :)

### DIFF
--- a/account_invoice_report_grouped_by_picking/README.rst
+++ b/account_invoice_report_grouped_by_picking/README.rst
@@ -90,6 +90,8 @@ Contributors
 
   * Manuel Regidor <manuelregidor@sygel.es>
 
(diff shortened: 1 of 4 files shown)


rafaelbn:
@Shide is this FIX similar to https://github.com/OCA/account-invoice-reporting/pull/318 ?

Ping @ferran-S73 😄 ❤️

pedrobaeza:
Yes, this seems to fix the same thing as #318. Please avoid all the noise of adding the contributor for a quick merge.

ferran-S73 (approved):

EmilioPascual (approved):

PR #324, opened 2024-06-07
[MIG][17.0] account_invoice_bank_details: Migration to 17.0

xaviedoanhduy:
- `account_bank_invoice_details` was introduced in v13 in PR https://github.com/OCA/account-invoice-reporting/pull/164
- It was not merged at that time, as it was considered potentially redundant with `account_payment_mode* ` modules
- After analysis, we still think that `account_bank_invoice_details` is a lighter alternative for a simple need that is not covered out- of-the-box: the automatic choice of the bank account based on the invoice currency


account_invoice_bank_details/readme/USAGE.md
```diff
@@ -0,0 +1 @@
+
```
gurneyalex: remove the empty file.

gurneyalex (approved):

account_invoice_bank_details/readme/USAGE.md
xaviedoanhduy: hi @gurneyalex, it's removed

Ricardoalso (approved):

PR #327, opened 2024-07-11
[15.0][FIX] account_invoice_report_grouped_by_picking: partial refund qty d…

christian-ramos-tecnativa:
…isplayed


christian-ramos-tecnativa:
This small fix solve an issue when refunding invoices using the reverse move option.

Tested branches 14.0, 15.0, 16.0, 17.0
Steps to replicate it:
1. Create and confirm a sale with one line but multiple qty
2. Deliver the products
3. Invoice the sale
4. Use the "Add credit note" option on the invoice
5. Modify the refunded qty to a lower one
6. Print the refund, the quantity displayed is always the delivered one

PR #329, opened 2024-07-16
[16.0][FIX] account_invoice_report_grouped_by_picking: Ensure exact search for tests instead of partial search

carlos-lopez-tecnativa:
When installed with other modules (e.g., sale_order_line_sequence), this commit makes the assertion more specific to avoid inconsistencies. sale_order_line_sequence adds a similar span (<span>S00001/1</span>), causing duplicate searches.

image

before

image
after, no error on test 

@tecnativa TT50196


pedrobaeza (approved):
Although this is a hackish patch, as parsing HTML of a report and looking for specific contents may be tricky due to other modules adding also contents there, but tests going to minimum existence is OK.

Fast-tracking for fixing the integration tests:

/ocabot merge patch

PR #337, opened 2024-10-17
[15.0][FIX] account_invoice_report_grouped_by_picking: Method abs is called when remaining_qty is near of zero

carlosdauden:
@Tecnativa TT51353

### DIFF
--- a/account_invoice_report_grouped_by_picking/models/account_move.py
+++ b/account_invoice_report_grouped_by_picking/models/account_move.py
@@ -2,8 +2,7 @@
 # Copyright 2018 Tecnativa - David Vidal
 # Copyright 2018-2019 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-import datetime
-from collections import OrderedDict
+from datetime import datetime
 
 from odoo import api, models
 from odoo.tools import float_round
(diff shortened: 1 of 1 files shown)


carlosdauden:
ping @sergio-teruel @chienandalu 

FWPort after merge

carlosdauden:
FWPort after merge

sergio-teruel (approved):

chienandalu (approved):

account_invoice_report_grouped_by_picking/models/account_move.py
```diff
@@ -101,45 +91,51 @@ def lines_grouped_by_picking(self):
                     }
                 )
                 continue
+
```
pedrobaeza: Please don't add empty lines inside a method, for both continuing with the existing style and as PEP8 recommendation.

carlosdauden: Changes done, but PEP8 allow blank lines in this cases:

> Extra blank lines may be used (sparingly) to separate groups of related functions. Blank lines may be omitted between a bunch of related one-liners (e.g. a set of dummy implementations).
> 
> Use blank lines in functions, sparingly, to indicate logical sections.
> 


https://peps.python.org/pep-0008/#blank-lines

carlosdauden: The main problem is that this method is too long and smaller methods should be created.
I discarded that change because all these methods are being declared in the account.move model.
The ultimate solution is to bring all this logic into an abstract_report, but I didn't want to do that in a stable release.
Let's keep that in mind for the migration to the new version.

pedrobaeza (approved):
/ocabot merge patch

PR #341, opened 2024-10-22
[15.0][FIX] account_invoice_report_grouped_by_picking: Recover default precision_rounding missed in previous PR

carlosdauden:
This is necessary when the invoice line does not have product.

FW-Port not needed because is OK in those versions

@Tecnativa TT51404

ping @sergio-teruel @CarlosRoca13


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #350, opened 2025-02-06
[17.0][FIX] account_invoice_report_grouped_by_picking: Display subtotals instead of totals

BT-etejeda:
When the same invoice line is linked to multiple pickings, the calculation shows the total price instead of the subtotal. That's what I'm trying to fix here. Besides the amount that is shown in the original Odoo invoice report is the subtotal.

ping @carlosdauden @pedrobaeza

### DIFF
--- a/account_invoice_report_grouped_by_picking/views/report_invoice.xml
+++ b/account_invoice_report_grouped_by_picking/views/report_invoice.xml
@@ -69,7 +69,6 @@
         </xpath>
         <xpath expr="//td/span[@t-field='line.price_subtotal']" position="before">
             <t t-set="line_subtotal" t-value="l.price_subtotal" />
-            <t t-set="line_subtotal" t-value="l.price_total" />
             <t t-if="lines_group['quantity'] != l.quantity" id="picking_subtotal">
                 <!-- Compute subtotal for that picking with discounts -->
                 <t


dalonsod:
Great! Today I've experienced such issue with a customer and tested with him the solution, seems to work, at least with an example

dalonsod (approved):
It looks good for an invoice with mix of lines belonging to a single picking and lines splitted in more than one :+1:

dalonsod:
@cramosh could you review?

dalonsod:
@BT-etejeda please note that this change alters aggregated subtotal calculation for a picking as well. Before this fix subtotal was shown tax included (even if it was properly calculated), now without taxes.

carlosdauden (approved):
Code LGTM

ofonseca-pyming (approved):
Functional LGTM

BT-etejeda:
> @BT-etejeda please note that this change alters aggregated subtotal calculation for a picking as well. Before this fix subtotal was shown tax included (even if it was properly calculated), now without taxes.

@dalonsod Yes, you're right. Thanks for the warning 🙂. Actually the subtotal of each subgroup continues to be the sum of the subgroup's amounts, only now that sum will actually correspond to the subtotal, as indicated by its label. :+1:

dalonsod:
Yes, this new behavior fits better to reality IMO

PR #356, opened 2025-02-26
[16.0][FIX] account_invoice_report_due_list: Define the right rate for the currency in test

victoralmau:
Define the right rate for the currency in test

In some cases (when installing some modules), the rate of the currency is different than expected, therefore, we remove the existing values and define the expected rate to validate the appropriate values.

@Tecnativa

### DIFF
--- a/account_invoice_report_due_list/tests/test_invoice_report_due_list.py
+++ b/account_invoice_report_due_list/tests/test_invoice_report_due_list.py
@@ -9,7 +9,7 @@ import babel
 from odoo import fields
 from odoo.tests import common, tagged
 from odoo.tests.common import Form
-from odoo.tools import posix_to_ldml, pycompat
+from odoo.tools import mute_logger, posix_to_ldml, pycompat
 
 
 @tagged("post_install", "-at_install")
@@ -108,7 +108,19 @@ class TestInvoiceReportDueList(common.TransactionCase):
         self.assertRegex(str(res[0]), date_due_format)
         self.assertRegex(str(res[0]), "75.0")
 
+    @mute_logger("odoo.models.unlink")
     def test_due_list_currency_extra(self, move_type="out_invoice"):
+        # We leave the rate of the extra currency at 1 so that the amounts to be
+        # checked are as expected.
+        self.env["res.currency.rate"].search(
+            [("currency_id", "=", self.currency_extra.id)]
+        ).unlink()
+        self.env["res.currency.rate"].create(
+            {
+                "currency_id": self.currency_extra.id,
+                "rate": 2,
+            }
+        )
         move_form2 = Form(
             self.env["account.move"].with_context(default_move_type=move_type)
         )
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Don't remove the extra argument that may be used by dependent modules.

account_invoice_report_due_list/tests/test_invoice_report_due_list.py
```diff
@@ -108,7 +108,20 @@ def test_due_list(self, move_type="out_invoice"):
...
         self.assertRegex(str(res[0]), "75.0")
 
+    @mute_logger("odoo.models.unlink")
     def test_due_list_currency_extra(self, move_type="out_invoice"):
+        # We leave the rate of the extra currency at 1 so that the amounts to be
+        # checked are as expected.
+        currency_extra_rates = self.env["res.currency.rate"].search(
+            [("currency_id", "=", self.currency_extra.id)]
+        )
+        currency_extra_rates.unlink()
```
pedrobaeza: No need of extra variable and less lines of code:

```suggestion
        self.env["res.currency.rate"].search(
            [("currency_id", "=", self.currency_extra.id)]
        ).unlink()
```

pedrobaeza (commented):
Anyway, this seems more something that the currency is not the same than the company currency. Have you explored to make sure the currency is the company's one in the invoices or whatever?

victoralmau:
> Anyway, this seems more something that the currency is not the same than the company currency. Have you explored to make sure the currency is the company's one in the invoices or whatever?

The problem is not the company's currency (in this test we want to validate the value that is shown when an invoice is made in a currency different from the company's), it is the value of the same that has changed, for that reason all the rates are removed and the appropriate value is set.

pedrobaeza:
Yeah, sorry, I saw it from mobile and didn't get full thing. Then for the test to serve for something, a different rate than 1 should be used and then get the amount converted to that rate to check that the code doing the conversion is working. If not, you don't know if the test is correct because is doing the conversion with rate 1, or no conversion is being done at all.

account_invoice_report_due_list/tests/test_invoice_report_due_list.py
```diff
@@ -108,7 +108,19 @@ def test_due_list(self, move_type="out_invoice"):
         self.assertRegex(str(res[0]), date_due_format)
         self.assertRegex(str(res[0]), "75.0")
 
+    @mute_logger("odoo.models.unlink")
     def test_due_list_currency_extra(self, move_type="out_invoice"):
+        # We leave the rate of the extra currency at 1 so that the amounts to be
```
pedrobaeza, line 113:
```suggestion
        # We leave the rate of the extra currency at 2 so that the amounts to be
```

victoralmau:
Reviewing in detail, it is not necessary to set a value (2 for example) for the opposite currency of the company because the value returned will be in the currency of the company; i.e. 50+150. 

I leave the PR in draft and I will continue investigating to apply the appropriate fix when I have more clear the reason and the solution, sorry.

victoralmau:
I see that the initial problem is when installing `account_payment_term_extension`, so probably the error is there, I continue analyzing.

victoralmau:
Not really necessary, explained at https://github.com/OCA/account-invoice-reporting/pull/356#issuecomment-2687186190.

PR #360, opened 2025-03-04
[FIX] account_invoice_report_product_sticker: allow accountants to print stickers

yajo:
Since the domain inside `_compute_sticker_ids` requires access to the `ir.model` model, only admins could print stickers.

Other users were getting an `AccessError`.

Now tests are executed with an accountant user. Running them like this without the fix triggers the error.

@moduon MT-9304


Shide (approved):

fcvalgar (approved):
LGTM, thank you @yajo

PR #369, opened 2025-06-02
[16.0][IMP] account_invoice_production_lot

edescalona:
@BinhexTeam

Modifying the display_type field comparison in the condition to display lots in the invoice report.


edescalona:
Ready for review

Issue #378, opened 2025-09-07
account_invoice_line_sale_line_position is not functional


nikolaszimmermann:
## Module
account_invoice_line_sale_line_position hasn't been ported correctly to Odoo 18.

## Describe the bug
Every invoice line should display an associated position from the corresponding sale line ids -- instead only the last line display a position, the all other lines show an empty position.

The fix is two-fold.

First problem: In my case the `if` condition `if record.display_type` is always true, since `display_type` is equal to `product` for regular products. Second, the `values = ...` block is incorrectly formatted (compare with the Odoo 13) version. It needs to be indented one more level, to be within the `for record in self` loop. Otherwise, when invoked with multiple account_move_lines at once, `record` will be equal to the _last line_ only.

```
diff --git a/account_invoice_line_sale_line_position/models/account_invoice.py b/account_invoice_line_sale_line_position/models/account_invoice.py
index 057e17c..9e232c9 100644
--- a/account_invoice_line_sale_line_position/models/account_invoice.py
+++ b/account_invoice_line_sale_line_position/models/account_invoice.py
@@ -25,11 +25,11 @@ class AccountMoveLine(models.Model):
     @api.depends("sale_line_ids.position")
     def _compute_position_formatted(self):
         for record in self:
-            if record.display_type:
+            if record.display_type != 'product':
                 record.position_formatted = ""
                 continue
-        values = [
-            val for val in record.sale_line_ids.mapped("position_formatted") if val
-        ]
+            values = [
+                val for val in record.sale_line_ids.mapped("position_formatted") if val
+            ]
 
-        record.position_formatted = "/".join(values)
+            record.position_formatted = "/".join(values)
```

## To Reproduce
**Affected versions**: 18

1. Install the module
2. Create sale order, add multiple products
3. Create invoice

**Expected behavior**
All lines should have a position, not only the last one.

Kimkhoi3010:
Hi @nikolaszimmermann, I have created the PR #379 to fix this.
Thanks for your time

PR #379, opened 2025-09-08
[18.0] [FIX] account_invoice_line_sale_line_position: compute position_formatted for all invoice lines

Kimkhoi3010:
The purpose of this PR is to fix the module `account_invoice_line_sale_line_position`, 
which is currently not functional in Odoo 18.

As already mentioned by @nikolaszimmermann in issue #378 ,
invoice lines should correctly display the position from their related sale lines.

### DIFF
--- a/account_invoice_line_sale_line_position/models/account_invoice.py
+++ b/account_invoice_line_sale_line_position/models/account_invoice.py
@@ -25,11 +25,10 @@ class AccountMoveLine(models.Model):
     @api.depends("sale_line_ids.position")
     def _compute_position_formatted(self):
         for record in self:
-            if record.display_type:
+            if record.display_type != "product":
                 record.position_formatted = ""
                 continue
-        values = [
-            val for val in record.sale_line_ids.mapped("position_formatted") if val
-        ]
-
-        record.position_formatted = "/".join(values)
+            values = [
+                val for val in record.sale_line_ids.mapped("position_formatted") if val
+            ]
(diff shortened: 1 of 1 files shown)


nikolaszimmermann:
Thanks for taking care!

account_invoice_line_sale_line_position/models/account_invoice.py
```diff
@@ -25,11 +25,10 @@ class AccountMoveLine(models.Model):
     @api.depends("sale_line_ids.position")
     def _compute_position_formatted(self):
         for record in self:
-            if record.display_type:
+            if record.display_type != "product":
```
nikolaszimmermann, line 28: I was unsure why this condition is relevant at all, in theory the code below should be sufficient. In my case, just dropping, it would work as well.

Maybe someone has insights why it was added in first place?

simahawk: It's like checking if you have `sale_line_ids` so in the end it shouldn't change much.
Probably it was fundamental in v14

nikolaszimmermann: Ok great. As long as the check is correct as-is we can leave it - up to you to decide.

rlimaeco (approved):

florentx (approved):

nikolaszimmermann:
Thanks, it's highly appreciated, time to remove my downstream override 👍

PR #380, opened 2025-09-20
[18.0][FIX] account_invoice_line_report: Make sure search view is empty

pedrobaeza:
Coming from previous versions, the search view has a value, so we should make sure that in this one, it's emptied for not having incorrect behaviors.

@Tecnativa


pedrobaeza:
Fast-tracking as this is something involving migrations and it's idempotent for the rest:

/ocabot merge patch

Issue #388, opened 2025-11-26
account_invoice_report_grouped_by_picking: for bom invoiced products, the product is counted X times


therbin200:
## Module
account_invoice_report_grouped_by_picking

## Describe the bug
When we invoice a bom product, it's counted as many times as we have components in the bom (at least with the phantom type)

## To Reproduce
**17.0, but possibly others too (tests have been processed on 17.0)**:

1. Create a Sale Order with a bom product with the bom type phantom
2. Confirm the delivery
3. Invoice the sale order

**Expected behavior**
The invoice report must include one time the finished product, not count it as many times as there is components in his bom in the delivery

**Additional context**
/

PR #389, opened 2025-11-26
[17.0][FIX] account_invoice_report_grouped_by_picking: Avoid quantity dupli…

therbin200:
…cation for kit products

Answers issue [#388](https://github.com/OCA/account-invoice-reporting/issues/388)


pedrobaeza (changes requested):
This can't be done in this module, as you are provoking that those not using `mrp` to fail. It should be an extra module. Check if it doesn't exist yet.

pedrobaeza (changes requested):
As commented.

therbin200:
OK, I've made an extra module here : #393

PR #393, opened 2026-01-09
[ADD] account_invoice_report_grouped_by_picking_mrp: Group invoice li…

therbin200:
…nes by picking with BOM awareness

This module groups pickings on invoices, taking existing BOMs into account.

    -If a picking contains products linked to a BOM, only the final product
invoice quantity is considered, instead of summing component quantities.


therbin200:
Answers issue [#388](https://github.com/OCA/account-invoice-reporting/issues/388)

PR #401, opened 2026-04-28
[FIX] account_invoice_production_lot: remove missing product_category_1 reference.

ysantiago:
In Odoo 19.0, the 'product.product_category_1' external ID has been removed in the base 'product' module data. This causes an error during the installation of the module or when loading test/demo records.

Since explicitly defining the 'categ_id' is not strictly required for this test record, this commit removes the field assignment entirely. The product will now use the standard default category, ensuring the module installs correctly without relying on deprecated data.


luisg123v (approved):
LGTM 👍

luisg123v:
@moylop260  could you review/merge, please?

CC @desdelinux

moylop260 (approved):

PR #421, opened 2026-06-08
[18.0][ADD] account_invoice_report_untaxed_unrounded

AungKoKoLin1997:
## Module

`account_invoice_report_untaxed_unrounded`

When an invoice uses **Round Globally** tax rounding with a zero-decimal currency (such as JPY), the per-line subtotal shown on the invoice report is rounded to the currency precision. The sum of these rounded line subtotals can then differ from the invoice untaxed total, which is computed on the unrounded amounts.

This module keeps the line subtotal and the invoice untaxed amount at full precision (`price_subtotal_unrounded` on `account.move.line` and `amount_untaxed_unrounded` on `account.move`, the latter equal to the sum of the former) and discloses them on the invoice report, formatted with a per-company display precision (*Invoicing Settings → Untaxed Unrounded Digits*) that you set finer than the currency. Rounding is therefore presentation-only; the stored fields stay exact and can be reused by other reports.

@qrtl QT6785


account_invoice_report_untaxed_unrounded/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,12 @@
+When an invoice uses **Round Globally** tax rounding with a zero-decimal
+currency (such as JPY), the per-line subtotal shown on the invoice report is
+rounded to the currency precision. The sum of these rounded line subtotals can
+then differ from the invoice untaxed total, which is computed on the unrounded
+amounts.
```
yostashiro: Move this to CONTEXT.md and comment that the behavior (on the invoice untaxed total) was introduced in 18.0.

yostashiro (commented):
Partial review.

account_invoice_report_untaxed_unrounded/readme/DESCRIPTION.md
AungKoKoLin1997: Done!

SanamiK (approved):
Functional review. LGTM.

Image

PR #424, opened 2026-06-24
[19.0][FIX] account_invoice_report_grouped_by_picking: convert grouped qty to invoice line UoM

rwagemans:
## Bug

`lines_grouped_by_picking` sums and subtracts the stock move quantity returned by `_get_signed_quantity_done` (expressed in the **move's UoM**, i.e. the product's stock UoM) against `line.quantity` / `remaining_qty` (expressed in the **invoice line UoM**), without converting between the two:

```python
remaining_qty = line.quantity            # invoice line UoM (e.g. 2 packs)
for move in line.move_line_ids:
    qty = self._get_signed_quantity_done(line, move, sign)  # move.quantity in stock UoM (e.g. 12 units)
    picking_dict[key] += qty             # printed in stock UoM
    remaining_qty -= qty                 # 2 - 12 = -10
...
if remaining_qty:                        # -10 is truthy
    lines_dict[line] = remaining_qty     # phantom remainder line
```

When the invoice line UoM differs from the product's stock UoM (e.g. selling in packs/pallets while stock is kept in units), two things go wrong:

1. The per-picking quantity is printed in the stock UoM instead of the invoice line UoM.
2. A bogus remainder line with a negative quantity is produced, because a stock-UoM quantity is subtracted from an invoice-line-UoM quantity.

It goes unnoticed in the common case where both UoMs are equal.

## Fix

`_get_signed_quantity_done` now converts the move quantity to the invoice line UoM before it is summed/subtracted. A regression test (`test_account_invoice_group_picking_uom_conversion`) covers a product sold per pack of 6 while delivered in units: exactly one group, expressed in packs, with no phantom remainder line.

## Test
`account_invoice_report_grouped_by_picking` test suite: 6/6 green.

### DIFF
--- a/account_invoice_report_grouped_by_picking/__manifest__.py
+++ b/account_invoice_report_grouped_by_picking/__manifest__.py
@@ -7,7 +7,7 @@
 {
     "name": "Account Invoice Grouped by Picking",
     "summary": "Print invoice lines grouped by picking",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.0.1",
     "category": "Accounting & Finance",
     "website": "https://github.com/OCA/account-invoice-reporting",
     "author": "Tecnativa, Odoo Community Association (OCA)",
--- a/account_invoice_report_grouped_by_picking/models/account_move.py
+++ b/account_invoice_report_grouped_by_picking/models/account_move.py
@@ -33,10 +33,20 @@ class AccountMove(models.Model):
         account_invoice_report_grouped_by_picking_sale_mrp module
         """
         if move.location_id.usage == "customer":
-            return -move.quantity * sign
-        if move.location_dest_id.usage == "customer":
-            return move.quantity * sign
-        return 0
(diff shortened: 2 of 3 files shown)


account_invoice_report_grouped_by_picking/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 {
     "name": "Account Invoice Grouped by Picking",
     "summary": "Print invoice lines grouped by picking",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.0.1",
```
pedrobaeza, line 10: Please don't change manually the module version. It's done on merge.

pedrobaeza (approved):

BhaveshHeliconia (approved):
LGTM!

pedrobaeza:
As no conflict has arisen with the module version change, let's merge with nobump:

/ocabot merge nobump

PR #429, opened 2026-09-02
[FIX][14.0] account_comment_template: fix the show the render of the …

hitrosol:
Placeholder expressions in the comment template were not being rendered correctly in the PDF report output. This PR fixes the rendering logic so expressions are properly evaluated and displayed.

Also added test cases to cover this scenario and prevent regression


hitrosol:
pre-commit error : "ModuleNotFoundError: No module named 'pkg_resources'"   ---
not related to this PR

